PR #6, opened 2017-03-21
Quickfixes (well, at first)

chrisj42:
Done, as much as possible now. I reimplemented spawners, and airwizard II (sort of; it's buggy, a bit), and dungeon chest keys. I also tweaked some game and debug settings, and added back an instructions menu. It's looking good...


chrisj42:
Oh, I added a few more things.

PR #7, opened 2017-03-30
Cleanup

chrisj42:
I've done a *lot* of fixing up -- including some new features. Main points I remember:
-heavily revised key input system; now allows shift/ctrl/alt, and multiple keys mapping to the same action.
-redid potions, and removed that huge annoying 100-line repetitive and virtually unnecessary if-else-if monster.
-removed / added / organized some title splashes
-lots of additional ideas in to-do list.


chrisj42:
oh, and simplified home system. Also added / fixed some things with the air wizard suit.

PR #10, opened 2017-05-02
Change some texts

Dannyiy:
Right, so, if you're on this page, it's for... uh... get a description. But uh... I don't have. So you can look at my commits... uh... one commit. I did only one. It's... short yes it's true.


chrisj42:
I might do this once I implement multiplayer mode.

PR #16, opened 2017-08-30
World seeds and fixed main method.

ChristofferHolmesland:
Added world seeds and fixed the main method so it can be used to generate worlds.


chrisj42:
Looks good!

PR #19, opened 2017-10-20
Localization and arrow changes

ChristofferHolmesland:
Added support for multiple languages. 
Changed arrow to reflect the suggestions in "Bugs and things to add" text file.


chrisj42:
So I sort of selectively implemented localization into the bug-fixes branch, since this is just messy and I'm not sure about arrows. If you want to do arrows, you can just open a new pull request with that one, I think.

PR #30, opened 2018-06-23
Added Spanish localization

PokuGamer:


TerisbackDev:
Does it work!

PokuGamer:
Yes, It does
image

PR #35, opened 2018-10-20
Fully fledged durability system

africalinguist:
has not been tested in every scenario, but it should work fairly well.


chrisj42:
I looked at it, and you did okay but not the best, and it would have broken in multiplayer. So I took the liberty of doing it myself.

PR #38, opened 2018-11-27
Added peaceful mode

africalinguist:
prevents monsters from spawning and you from losing hunger, tested on both client and server, so it should work.


africalinguist:
I don't really feel like doing this.

chrisj42:
heh

PR #45, opened 2019-03-23
Arrow hit limit

africalinguist:
Adds a limit to the number of entities an arrow can hit before disappearing. Also reduces the damage for hits after the first.

### DIFF
--- a/ideas/ideas for feature updates.txt
+++ b/ideas/ideas for feature updates.txt
@@ -43,10 +43,6 @@ Allow clients to add a port number to the end of the ip with ":". for example: 1
 
 ----
 
-make arrows go away after hitting a certain number of enemies, or dealing a certain amount of damage
-
-----
-
 make creeper holes like TNT, just only one block. Though I should still have the thing where it hurts entities based on distance.
 
 ----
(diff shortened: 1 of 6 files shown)


africalinguist:
like that?

chrisj42:
Did you test it? think about what happens on a server. If you return false from doHurt for all remote players, then the arrow will never actually go away if it just hits players. But if you return true, then it will damage every frame.
So you see, there's actually a bigger problem here. Clients should not be dealing with damage to themselves, because then it messes up the server, and besides, it also opens up the ability to cheat.
And then, if you look at arrows, you see that they are updated on both the client and server, and that doesn't make total sense either. After all, then the client would track all arrow hits, and would remove the arrow, but the server would not remove it, and would continue until it hit 4 non-player mobs, resulting in apparent "ghost damage" in the client side.
Ideally, they would all be updated on the server, and the only things updated on the client are those that the server doesn't track, i.e. the particle subclasses. However, the player is updated partially on the client to mitigate lag.

So essentially, I can't accept this until you have a way of checking if the player actually got hurt on the server, so the ghost damage doesn't happen.

PR #49, opened 2019-03-24
Fixed black gaps between text in "Click to focus" message

africalinguist:
title says it all, also added one tile to icons.png


chrisj42:
looking good

PR #50, opened 2019-03-24
Arrow hit limit

africalinguist:
lol accidentally closed this


africalinguist:
I have found that I don't know enough about the play hit code and how it works to implement this, closing.

PR #52, opened 2019-03-30
Add escape potion

africalinguist:
commits say it all


chrisj42:
so the main problem with this is that the client MUST have the level above loaded, otherwise it crashes. Besides that, it uses an outdated reference to the level, which it shouldn't be doing ever. It probably shouldn't even cache old levels at all imo. Anyway in addition, both the client and the server try to position the player, which is odd.

chrisj42:
alright, now we hopefully are good.

PR #53, opened 2019-04-28
Made flowers able to spread grass

africalinguist:
Flowers now spread grass at a higher rate than normal grass.


chrisj42:
I mean, it looks like you literally just copied the code from GrassTile.java... which will work, I just highly despise copied code. I suppose it's all you can do though.

PR #58, opened 2019-05-04
Chests now actually save tool durability

africalinguist:
this is being exploited on servers to repair items for free.


chrisj42:
One thing I just realized; chests save tool durability, if only in a bit of a silly way; you use a separate delimiter ("." instead of ";"), and yet you rely on the fact that the item doesn't have a ";" section as well. Since you're already requiring them to be mutually exclusive, why not go one step further and reduce the data format complexity by just using the same symbol and checking the class when loading it back? Better yet, have the Item subclass determine what to do with the data, and let the data be a generic string who's content Load.java doesn't worry about. That would be nice in my opinion.

Along that vein, what do you think about having item stackability depend on whether the item has data? After all, stacks are great, and the only instance where an item can't be made part of a stack is when the item has individual data like durability. 

Part of the reason I suggest this change is because it means that you don't have to worry about as many things as a developer. For instance, you forgot about CHESTIN/CHESTOUT network requests and how data is handled there; they actually do it in a way that I think should be adopted by the Save/Load code. They use item.getData() which contains the name, stack size, and durability, whatever there is. Not that there's ever durability and a stack size at the same time. Anyway, point is, I think making the item class itself do the data parsing makes more sense than having an external class do it. It allows for flexible implementation and better encapsulation of class fields and such.
So what do you think?

chrisj42:
In hindsight this is definitely overkill for the problem at hand.

But it's for the good of the codebase. :D

chrisj42:
alright this should work.

PR #60, opened 2019-05-04
Remove unneeded code

africalinguist:
(a.k.a remove all the commented code (a.k.a reduce the number of annoying warnings my code checker is shooting at me))


chrisj42:
So while I get the point with this, and I think some should be removed, idk if they all should..? I mean they probably all should tbh, but also, like... sigh idk. I'd have to look through them in detail. I'll do that soon.

PR #62, opened 2019-05-12
Clarify hovered tile

africalinguist:
read the commits


chrisj42:
super picky nitpick here, but shouldn't the overlay be above the "swimming" oval thing? Since it's technically supposed to be the water too. Also, since it's rendered with the player, whether or not it renders on top of other entities may vary. But I suppose that's really not an issue.

PR #64, opened 2019-05-18
Cloud cacti

africalinguist:
added some functionality to cloud cacti
(made them able to drop from the tile, crafted into cloud dust, etc.)

more functionality for the cloud dust will be coming in a future pr


chrisj42:
I think this should be part of 2.0.6

PR #65, opened 2019-05-19
Fixed the bug with terrain vertically shredding!

africalinguist:
Hopefully this fixes it once and for all.


chrisj42:
yep, here's hoping it works.

PR #66, opened 2019-05-28
Small bugfixes

africalinguist:
fixed a few small things that I noticed.


chrisj42:
While I think this is good, shouldn't we also check other tiles for creative mode breakability? For example, it seems that the HardRockTile just pretends that a creative mode player successfully attacked, without actually doing anything... what's with that? Also, keep in mind that most tiles check for creative mode players and automatically adjust the damage to full health in the final hurt method that doesn't specify source. So if unbreakable tiles just always called hurt with zero-damage, and displayed their message only if not in creative mode, I think that would be a good way to handle it.

chrisj42:
Upon reviewing the code again, it seems that the way attacking works is that, for every target tile, the interactOn of the item is called first, then the interact method of the tile is called, and then finally, if neither of those return true and the item can attack, then the hurt method is called on the tile.
Based on this, it would seem that creative mode players would not be able to use a tool on hard rock because it returns true from interact without actually causing damage. Very strange. I just checked, and indeed, attacking hard rock with a pickaxe in creative has no response.

PR #70, opened 2019-06-01
Made some small changes

africalinguist:


chrisj42:
wait, why did you change the indexing of the random splash? Isn't it never going to choose the first one anymore?

MaxACartwright:
v

africalinguist:
It's so that the splashes for certain dates will never be picked if it's not that date

chrisj42:
ah I see, it didn't seem that way in the splashes commit, but I see now.

PR #73, opened 2019-06-08
Major bugfixes

africalinguist:
lots of important stuff that should be fixed right away


chrisj42:
I like it! You really seem to be getting the hang of the code base. Make sure to consider multiplayer functionality where it applies, however. It's pretty tricky. This seems good though. Can you confirm the creeper fix works on SP and MP modes?

africalinguist:
I can confirm it works,

chrisj42:
alright then.

PR #83, opened 2019-06-14
Fishing overhaul

africalinguist:
large changes to fishing


africalinguist:
it seems to work, but it doesn't render the fishing rod in front of the player when fishing on servers

africalinguist:
I would say that fishing rods perform a very different function, and that it's more convenient to not have to make a bunch of exceptions for if it's a fishing rod in ToolItem.java

PR #86, opened 2019-06-17
More structures

africalinguist:
Added more structures to the game and allowed you to place floors on cloud tiles


africalinguist:
this had a lot of "small fixes" commits, but I'm happy with it now

PR #95, opened 2019-07-09
added simplified chinese localization

muirmok:


ChristofferHolmesland:
You need to add the chinese characters to the icons.png file before they can be rendered in the game.

Makkkkus:
This should be closed

PR #99, opened 2019-07-20
Maps, books, and jungles

africalinguist:
This is the "Bamboo update" from the devlog discord channel, but without the editable books, I felt that the saving and loading of book content wasn't really ready. (and it would introduce a bunch of server problems)


africalinguist:
I've spent a little while squashing some of the bugs this introduced, I just want a quick look from Chris J and I'll merge

africalinguist:
there have been too many changes since I started this

PR #100, opened 2019-07-20
Added annotations library for Eclipse users.

Makkkkus:


ChristofferHolmesland:
When you want to add a file that is ignored by the .gitignore file, you should use the following command instead of modifying the ignore file `git add -f file`.

I'm not sure if adding another dependency to the repository is the right choice, as you typically only want to store files directly related to your software in them. A tool like Maven is probably a better option.

It is also important to be careful when using libraries with different licences. The project currently contain software with the following licences: Apache 2.0, BSD 3-Clause, JSON, EPL-1 and MIT. The project itself is licences as GPL-3. I'm not sure what the ramifications of using all of these together are, or if it is legal. E.g. has all of the NOTICE files been kept?

Makkkkus:
Ok, I'll close the pull request.
Thanks for the feedback anyways.

PR #103, opened 2019-07-31
Fixed a sound bug and added a new splash text.

Makkkkus:
also added a new debug message at line 51 in the Game class.


africalinguist:
I'm fine with these changes, but what was the sound bug?

Makkkkus:
There is no sound when closing the inventory, but there is a sound when closing the crafting menu.

africalinguist:
ah, ok

PR #112, opened 2019-12-28
Renamed mispelled function

jaredible:


Makkkkus:
Gamer moment for sure!

PR #145, opened 2021-02-07
Save spanish_es.mcpl in UTF-8 encoding

alice39:
accented words can't be rendered due that file wasn't saved in UTF-8 encoding


Makkkkus:
Fixed in my fork

alice39:
Closing because fixed previously in #146

PR #150, opened 2021-02-26
Add item "Natural rock" in creative mode

TheBigEye:
User can add rock, it is useful as once rock is broken in creative mode, there is no way to regenerate it.


TheBigEye:
Ready to pull

Makkkkus (approved):
Pretty good. Version should be -dev1, not -dev2 though.

TheBigEye:
ok

chrisj42:
so about that dev1... haha

chrisj42:
also you're going to have to back-merge with the new master, and fix all your edits @TheBigEye since there have been some changes.

chrisj42:
Oh, it seems that a lot of these edits are random QoL stuff and code style too... better to keep those things in a separate request anyway.

TheBigEye:
I will do everything again

PR #153, opened 2021-03-06
Save and load the used seed in generating a world

alice39:
This will simply save and load a seed in level files to show it in debug info [ F3 ]

If it was a previous world before that 2.0.7-dev2 or if it's an online world it'll show 0 by default


chrisj42:
I like it, handy addition.

PR #173, opened 2021-08-08
Use Gradle to build game

Makkkkus:


Dioarya:
I don't know how these gradle stuff works, so may I ask why did you put `.gradle` in the .gitignore?

Makkkkus:
> 
> 
> I don't know how these gradle stuff works, so may I ask why did you put `.gradle` in the .gitignore?

.gradle is a folder which is made up of build artifacts; Not stuff you want picked up by git.

Dioarya:
I think you accidentally turned every line ending from CRLF into LF...

PR #190, opened 2021-08-30
Ability to preview Skins in menu!

Litorom:


src/main/java/minicraft/screen/SkinDisplay.java
```diff
@@ -149,14 +148,53 @@ public void render(Screen screen) {
...
-		int w = 15;
+		int w = 16;
 		int xo = (Screen.w - w * 8) / 2;
 		int yo = 28;
 
-		// TODO: Replace the logo with a preview of the skin to be selected
 		for (int y = 0; y < h; y++) {
 			for (int x = 0; x < w; x++) {
-				screen.render(xo + x * 8, yo + y * 8, x + y * 32, 0, 3); // Skin preview?
+				if (selected == 0) {
```
Makkkkus, line 157: Find out a way to make this more dynamic. Start by making this a for loop.

Litorom: You mean to simplify it so it's smaller?

Makkkkus: Exactly!

Makkkkus (dismissed):
Commit one good, commit two bad

PR #197, opened 2021-09-01
end update idea

RandomTNT:


Makkkkus:
I do suggest creating an issue in the issue tab instead.

El-Virus:
Yeah use the issue tab, to create suggestions, the pr tab is exclusive for submitting code.

PR #206, opened 2021-09-25
Menu Split and Fixes

Litorom:
#205 fixed


Litorom:
Sorry it glitched

src/main/java/minicraft/screen/OptionsMainMenuDisplay.java
```diff
@@ -0,0 +1,30 @@
+package minicraft.screen;
```
Makkkkus, line 1: Why does OptionsDisplay still exist?

Makkkkus (changes requested):
Changes are good, but you need to remove OptionsDisplay so we don't have three settings menu classes.

This is a nice little pull request, though.

src/main/java/minicraft/screen/OptionsMainMenuDisplay.java
Litorom: I said in #205, that I kept it just in case.

Makkkkus (commented):
Changes are good; will test after school is done

Makkkkus (approved):
Approved (it didn't remove the review last time)

Litorom:
MERGE!!

PR #212, opened 2021-11-06
Player.java updated

rockedsocks:
The player should now spawn on a tile ~5 tiles from spawn whenever they don't have a bed.
Also the `respawn` function now returns `void`. Used to return true, but idk if that breaks anything.


Makkkkus:
Remove the whitespace changes please

PR #215, opened 2021-11-07
Random Respawn when without Bed

rockedsocks:
Player now semi-randomly respawns when there is no bed.

i have had my will finally broken and have resorted to using the built in github.com editor

:(


Makkkkus:
👏🏻 👏🏻 👏🏻

PR #220, opened 2021-11-28
Fix for illogical Obsidian and additional tiles

### LINKED ISSUE #219: [FEATURE] Adding raw obsidian for logic reasons and corresponding Decor Tiles
**Additional Info (12/07/2021):**
Add DecorTiles for more tile diversity. 

The way water and lava make obsidian “planks” makes no sense and is illogical.

I recommend changing the texture and name to something like “Raw Obsidian” can be a new tile which in this case can be mined and used to make Obsidian Planks.

**Describe alternatives you've considered**
An alternative that requires less coding is renaming it to obsidian and changing the texture.

Litorom:
Fix for illogical Obsidian and additional tiles such as Stone (Raw), Ornate Stone, and Obsidian Tiles.
Fixes #219.


Makkkkus:
Will have to bugfix this, but it seems to work.

Issue #218, opened 2021-11-26
Typing certain chars in Inventory Searcher causes crash


Litorom:
**Describe the bug**
Typing certain chars such as “(“, “[“, “{“ into Inventory Searcher causes a crash

**Desktop (please complete the following information):**
 - OS: Windows 11
 - Version 2.0.7+

PR #221, opened 2021-11-28
Fix that certain characters crash Inventory Search

alice39:
This is because of String#replaceAll uses regex, which it is asking
some regex pattern thereby now it's changed it to String#replace because
it doesn't use regex, avoiding this crash when typing characters like
"{", "[" or "("

This fixes related issue #218
image


Litorom:
Fixes #218

Issue #211, opened 2021-10-30
Game crashes when pressing ESC after deleting world


SashaTalk:
Game crashes when jumping to previous menu after world deleting

**Steps to reproduce the behavior:**
1. Start game
2. Select "Singleplayer"
3. Press "D"
4. Press "Enter" to delete highlighted world
5. Press "Enter" again to confirm deleting
6. Press "ESC"
7. Press "Enter"

**Expected behavior**
Game crashed after pressing "Enter"

**Video**
https://user-images.githubusercontent.com/76222964/139535643-43937bff-3640-4f4a-8c2f-f6c7e89a7093.mp4

**Desktop:**
 - OS: Microsoft Windows 10 21H1 (10.0.19043.1288)
 - Version 2.0.7

Makkkkus:
I cannot replicate the problem.

lino-levan:
I was able to replicate the error. May look into fixing this is a little bit.

https://user-images.githubusercontent.com/11367844/143929670-099571a3-08c4-446f-bd8b-0aa6196faec9.mp4

Makkkkus:
Aha

Litorom:
This code appears to be what the error shows, line 77.

for (File file : list) {
	file.delete();
}
world.delete();

lino-levan:
Can confirm. Replacing it with:

```java
if(list != null) {
  for (File file : list) {
	  file.delete();
  }
}
```

Fixes the crash but the real question is why does pressing "esc" return to the delete world menu?

Litorom:
This is where the “ESC” key usage is used.
https://github.com/chrisj42/minicraft-plus-revived/blob/master/src/main/java/minicraft/screen/Display.java
Brings you back to the previous menu. This must be changed so you can’t do that in this menu.

lino-levan:
@Makkkkus Should this issue be closed?

PR #223, opened 2021-11-30
Fix parent when going back to World Select Display

alice39:
When action is done either copy, rename or delete
this goes from World Select Display to World Edit
Display thereby that minicraft.core.Updater class
set the new parent to new menu that is World Edit
Display, but when it goes from World Edit Display
to World Select Display, minicraft.core.Updater
class initializes it with World Edit Display so
when user tries to exit from menu, it'll go back
again to previous World Edit Display. To avoid
this bug, World Select Display will select parent
correctly when editing process started if given
conditions are true that means that will pick
the select's parent from edit's parent what it'll
result as the initial parent

This fixes related issue #211
also I'm not sure if this might have another good fix for this bug but this is what it reaches first to me

### DIFF
--- a/src/main/java/minicraft/screen/WorldSelectDisplay.java
+++ b/src/main/java/minicraft/screen/WorldSelectDisplay.java
@@ -110,7 +110,14 @@ public class WorldSelectDisplay extends Display {
 	
 	@Override
 	public void init(Display parent) {
-		super.init(parent);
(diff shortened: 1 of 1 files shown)


Litorom (commented):
Seems to work, cannot test at time of this comment.

lino-levan:
I imagine that this works but I feel like this should be more configurable? I can't think of a better way of doing it off the top of my head though.

PR #227, opened 2021-12-03
Fixed issue where entity despawned in player's sight

### LINKED ISSUE #148: Animals dissapear
Animals disappear after a while regardless if it is in view or not. Functionality should be added to keep animals, and de-spawning should be prevented while in view.

lino-levan:
Fixed #148


Makkkkus:
Very good

PR #233, opened 2021-12-08
Cleaned up code

lino-levan:
- Removed a lot of unused imports
- Cleaned up some code that was technically deprecated

### DIFF
--- a/src/main/java/minicraft/core/io/Localization.java
+++ b/src/main/java/minicraft/core/io/Localization.java
@@ -48,7 +48,7 @@ public class Localization {
 		if (key.matches("^[ ]*$")) return key; // Blank, or just whitespace
 		
 		try {
-			double num = Double.parseDouble(key);
+			Double.parseDouble(key);
 			return key; // This is a number; don't try to localize it
 		} catch(NumberFormatException ignored) {}
 		
--- a/src/main/java/minicraft/entity/furniture/Furniture.java
+++ b/src/main/java/minicraft/entity/furniture/Furniture.java
@@ -52,7 +52,7 @@ public class Furniture extends Entity {
 	@Override
 	public Furniture clone() {
 		try {
-			return getClass().newInstance();
+			return getClass().getDeclaredConstructor().newInstance();
 		} catch (Exception ex) {
 			ex.printStackTrace();
 		}
--- a/src/main/java/minicraft/entity/mob/RemotePlayer.java
+++ b/src/main/java/minicraft/entity/mob/RemotePlayer.java
@@ -7,7 +7,6 @@ import org.jetbrains.annotations.Nullable;
 
 import minicraft.core.Game;
 import minicraft.core.io.InputHandler;
-import minicraft.entity.ClientTickable;
 import minicraft.entity.Direction;
 import minicraft.entity.Entity;
 import minicraft.entity.furniture.Bed;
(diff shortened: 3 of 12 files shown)


src/main/java/minicraft/entity/furniture/Furniture.java
```diff
@@ -52,7 +52,7 @@ public Furniture(String name, Sprite sprite, int xr, int yr) {
 	@Override
 	public Furniture clone() {
 		try {
-			return getClass().newInstance();
+			return getClass().getDeclaredConstructor().newInstance();
```
Makkkkus, line 55: Scary code

src/main/java/minicraft/core/io/Localization.java
```diff
@@ -48,7 +48,7 @@ public static String getLocalized(String key) {
 		if (key.matches("^[ ]*$")) return key; // Blank, or just whitespace
 		
 		try {
-			double num = Double.parseDouble(key);
+			Double.parseDouble(key);
```
Makkkkus, line 51: Is this necessary?

src/main/java/minicraft/entity/mob/RemotePlayer.java
```diff
@@ -18,7 +17,7 @@
 import minicraft.level.Level;
 
 /** This is used for players in multiplayer mode. */
-public class RemotePlayer extends Player implements ClientTickable {
+public class RemotePlayer extends Player {
```
Makkkkus, line 20: Are you sure this won't break something?

src/main/java/minicraft/saveload/LegacyLoad.java
```diff
@@ -104,7 +104,6 @@ public void loadFromFile(String filename) {
 			br = new BufferedReader(new FileReader(filename));
 			
 			String curLine;StringBuilder total = new StringBuilder();
-			ArrayList<String> curData;
```
Makkkkus, line 107: Please don't touch LegacyLoad. It is called that for a reason.

src/main/java/minicraft/screen/TitleDisplay.java
```diff
@@ -23,7 +21,6 @@
 import minicraft.level.Level;
 import minicraft.screen.entry.BlankEntry;
 import minicraft.screen.entry.LinkEntry;
-import minicraft.screen.entry.ListEntry;
 import minicraft.screen.entry.SelectEntry;
```
Makkkkus, line 24: Good

src/main/java/minicraft/core/io/Localization.java
lino-levan: Technically not _necessary_, but removing unnecessary variable declarations/assignments is probably a good idea.

src/main/java/minicraft/entity/mob/RemotePlayer.java
lino-levan: Player already implements ClientTickable, so this is unneeded verbosity imo

src/main/java/minicraft/saveload/LegacyLoad.java
lino-levan: Fair, I will remove this

src/main/java/minicraft/entity/furniture/Furniture.java
lino-levan: How so? `getClass().newInstance()` skips compile-time checks of constructors. I would rather get a compile error (with `getClass().getDeclaredConstructor().newInstance()`) than a runtime error from this.

src/main/java/minicraft/screen/TitleDisplay.java
lino-levan: :)

src/main/java/minicraft/entity/furniture/Furniture.java
Makkkkus: Thank you for explaining

Issue #241, opened 2021-12-13
Remove green dots from dirt dungeon tile texture


pelletsstarPL:
![obraz](https://user-images.githubusercontent.com/54317957/145837731-36e4c524-e059-40fc-be8d-b4257d78c399.png)
Those green dots make it look weird tbh
On screenshot you have how you can improve it.

TheBigEye:
🤔

TheBigEye:
I think this should be closed

PR #243, opened 2021-12-26
Remove green dots from dirt dungeon tile texture

TheBigEye:
Removed green dots and updated the texture of the dungeon dirt tile:

image

 This solves #241


TheBigEye:
I can't believe it was very easy ._.

Makkkkus:
👍🏻

Issue #200, opened 2021-09-07
FPS doesn't match settings


Makkkkus:
When I set my FPS to 60 in the settings the actual FPS is closer to 47.

Litorom:
Issue #200!! We have 200!!!

lino-levan:
> When I set my FPS to 60 in the settings the actual FPS is closer to 47.

How did you test this? Is there an in-game counter for FPS hidden somewhere?

Makkkkus:
> How did you test this? Is there an in-game counter for FPS hidden somewhere?

F3

lino-levan:
Don't know why I didn't think of that

PR #252, opened 2022-01-10
Fixed FPS not matching in-game cap

lino-levan:
Does anyone know why there was the 2 millisecond pause in the main game loop? I haven't experienced any issues from removing it and now the game runs perfectly at the fps cap.

### DIFF
--- a/src/main/java/minicraft/core/Initializer.java
+++ b/src/main/java/minicraft/core/Initializer.java
@@ -118,12 +118,6 @@ public class Initializer extends Game {
 				unprocessed--;
 			}
 			
-			try {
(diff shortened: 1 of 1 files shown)


lino-levan:
I should mention that this closes #200.

Makkkkus:
Really good work!

alice39:
This change sadly makes Minicraft use more CPU usage from 30% to 70% in my laptop that is not that good of specifications, but yeah it solves FPS issue

lino-levan:
We really need to work on a better game loop I would say... we are literally just running a while(true) loop essentially in the background as our main game loop.

PR #259, opened 2022-01-16
Achievable Achievements Part 1

Litorom:
Included
+Added Debug for when an Achievement's data is changed (either achieve or unachieve)
+Made "Fashion Show", "Defeat... the air?", "An airy feat!", and "Demolition Demo" all achievable.

### DIFF
--- a/src/main/java/minicraft/screen/SkinDisplay.java
+++ b/src/main/java/minicraft/screen/SkinDisplay.java
@@ -103,6 +103,9 @@ public class SkinDisplay extends Display {
 
     @Override
     public void onExit() {
+		// Achieve Fashion Show:
+		AchievementsDisplay.setAchievement("minicraft.achievement.skin", true);
+
 		// Save the selected skin.
 		new Save();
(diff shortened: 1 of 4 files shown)


src/main/java/minicraft/screen/SkinDisplay.java
```diff
@@ -103,6 +103,9 @@ public SkinDisplay() {
 
     @Override
     public void onExit() {
+		// Achieve Fashion Show:
+		AchievementsDisplay.setAchievement("minicraft.achievement.skin", true);
```
Makkkkus, line 107: Shouldn't this be achieved when opening the menu?

Litorom: Why? If you just open the menu you are only viewing skins not causing an update.
This is also better if you don’t want to change your skin just select the same exact skin.

Makkkkus: 👍🏻

PR #269, opened 2022-02-10
Increase vanilla ID limit to 32768

PseudoDistant:
Still no method of converting old torch tiles in old worlds, since torch tiles now start after 16384.


PseudoDistant:
Never mind, seems it will fetch from the name. :/

PR #282, opened 2022-02-26
Update with not using a very long Array

BenChueng0422:
I was just copy-pasting from mine, so I have to resolve that first XD


Makkkkus:
Redundant

PR #286, opened 2022-03-09
Book Error fix

### LINKED ISSUE #285: Books that are fished have error message
Whenever you fish a book and read it it says "This is an empty book, this should probably not happen"
image

Litorom:
Didn’t test but should work

Closes #285.


Makkkkus:
Sorry for actually not merging fast enough

Issue #277, opened 2022-02-23
Implement support for different screen resolutions


JamesTDG:
For quite a while, Minicraft has been stuck in 4:3, which is almost a square when it comes to resolution. You can go into fullscreen with the application, but the used visual space is still 4:3, I wish for this to be changed, since I want to be capable of seeing more of my world at once. Here's a set of recommended changes to help improve this:

= **>Vital<** Implement support for multiple different resolutions

= _>Optional<_ Make it possible to zoom in and out, possible limitations can include a maximum of 2x zoomed in and a maximum of ~4x zoomed out. Would be recommended to have zooming disabled in places without any natural light.

PR #301, opened 2022-04-07
Added multiple screen resolutions

lino-levan:
Is the start of dealing with #277. We probably need to rethink some of the menus but this works good enough as a start for now.

The default of 4x3 makes sense until we adjust menus for both 4x3 and 16x9.

The "Native" screen resolution option adjusts the screen resolution to that of the main monitor. On a Mac, it'd be 16x10. On most monitors, it'd be 16x9.

### DIFF
--- a/src/main/java/minicraft/core/Renderer.java
+++ b/src/main/java/minicraft/core/Renderer.java
@@ -10,6 +10,7 @@ import java.util.ArrayList;
 import java.util.Map;
 import java.util.Objects;
 
+import minicraft.core.io.Settings;
 import minicraft.entity.furniture.Bed;
 import minicraft.entity.mob.Player;
 import minicraft.gfx.Color;
@@ -35,8 +36,9 @@ import javax.imageio.ImageIO;
 public class Renderer extends Game {
 	private Renderer() {}
 	
-	public static final int HEIGHT = 192;
-	public static final int WIDTH = 288;
+	public static final int MAX_SIZE = 288;
+	public static int HEIGHT;
+	public static int WIDTH;
 	static float SCALE = 3;
 	
 	public static Screen screen; // Creates the main screen
@@ -78,7 +80,36 @@ public class Renderer extends Game {
 		return new SpriteSheet[] { itemSheet, tileSheet, entitySheet, guiSheet, skinsSheet };
 	}
 
-	static void initScreen() {
+	public static void setScreenResolution() {
+		int width = 0;
+		int height = 0;
+
+		switch ((String) Settings.get("screenresolution")) {
+			case "Native":
+				java.awt.Dimension size = java.awt.Toolkit.getDefaultToolkit().getScreenSize();
+				width = (int) size.getWidth();
+				height = (int) size.getHeight();
+				break;
+			
+			case "4x3":
+				width = 4;
+				height = 3;
+				break;
+
+			case "16x9":
+				width = 16;
+				height = 9;
+				break;
+			default:
+				break;
+		}
+
+		double s = Math.min((double) Renderer.MAX_SIZE/width, (double) Renderer.MAX_SIZE/height);
+		Renderer.HEIGHT = (int) (height * s);
+		Renderer.WIDTH = (int) (width * s);
+	}
+
+	public static void initScreen() {
 		image = new BufferedImage(WIDTH, HEIGHT, BufferedImage.TYPE_INT_RGB);
 		pixels = ((DataBufferInt) image.getRaster().getDataBuffer()).getData();
 
--- a/src/main/java/minicraft/core/io/Settings.java
+++ b/src/main/java/minicraft/core/io/Settings.java
@@ -13,6 +13,7 @@ public class Settings {
 	private static final HashMap<String, ArrayEntry> options = new HashMap<>();
 	
 	static {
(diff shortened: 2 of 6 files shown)


Makkkkus:
This is aspect ratio, not screen resolution.

lino-levan:
this is a great point, let me fix this

lino-levan:
@Makkkkus does this look good?

src/main/java/minicraft/core/Renderer.java
```diff
@@ -78,7 +80,36 @@ private Renderer() {}
...
+			case "16x9":
+				width = 16;
+				height = 9;
+				break;
+			default:
+				break;
+		}
+
+		double s = Math.min((double) Renderer.MAX_SIZE/width, (double) Renderer.MAX_SIZE/height);
+		Renderer.HEIGHT = (int) (height * s);
```
Makkkkus, line 108: I can't think how this would work in my head alone. Does this secure that the scale of the window is the same as before?

src/main/java/minicraft/core/io/Settings.java
```diff
@@ -13,6 +13,7 @@
 	private static final HashMap<String, ArrayEntry> options = new HashMap<>();
 	
 	static {
+		options.put("aspectratio", new ArrayEntry<>("Aspect Ratio", "4x3", "16x9", "Native"));
 		options.put("fps", new RangeEntry("Max FPS", 10, 300, getRefreshRate())); // Has to check if the game is running in a headless mode. If it doesn't set the fps to 60
```
Makkkkus, line 17: Put it at the end of this list.

src/main/java/minicraft/screen/OptionsMainMenuDisplay.java
```diff
@@ -20,6 +24,14 @@ public OptionsMainMenuDisplay() {
         );
     }
 
+    @Override
+	public void render(Screen screen) {
+		super.render(screen);
+        
+		// Warning users about screen resoluton requiring a restart
+		Font.drawCentered("Some settings may require a restart", screen, Screen.h-20, Color.RED);
+    }
```
Makkkkus, line 33: This isn't good enough. Ideally we would do the same as with the dialog box that appears when resetting keybinds.

src/main/java/minicraft/screen/OptionsMainMenuDisplay.java
```diff
@@ -20,6 +24,14 @@ public OptionsMainMenuDisplay() {
         );
     }
 
+    @Override
+	public void render(Screen screen) {
+		super.render(screen);
+        
+		// Warning users about screen resoluton requiring a restart
```
Makkkkus, line 31: Would be nice to make this work without restarting the game.

Makkkkus (changes requested):
Good idea, but I don't think this is safe enough yet.

This might also break checking for entities on screen, and other quirks like that.

src/main/java/minicraft/screen/OptionsMainMenuDisplay.java
lino-levan: It would be, but it would require messing with a LOT of init code.

It would at minimum require another call to
```java
Initializer.createAndDisplayFrame();
// and
Renderer.initScreen();
```

But this would just create another display without closing the old one. It may be worth looking into in the future, but unless it's a dealbreaker for you, I don't think it's worth adding complexity for this.

Makkkkus: It's not a dealbreaker. I just wanted to make the point.

src/main/java/minicraft/core/Renderer.java
```diff
@@ -78,7 +80,36 @@ private Renderer() {}
 		return new SpriteSheet[] { itemSheet, tileSheet, entitySheet, guiSheet, skinsSheet };
 	}
 
-	static void initScreen() {
+	public static void setAspectRatio() {
+		int width = 0;
```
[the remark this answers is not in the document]
lino-levan: This has to be done before `Initializer.createAndDisplayFrame();`, I'm about to push a commit which fixes these comments where I just add it to that method instead.

lino-levan:
@Makkkkus does this look good?

src/main/java/minicraft/screen/TitleDisplay.java
```diff
@@ -55,7 +54,9 @@ public TitleDisplay() {
 								new SelectEntry("Credits", () -> Game.setMenu(new BookDisplay(BookData.credits)))
 						).setTitle("Help").createMenu()))
 				),
-			new SelectEntry("Quit", Game::quit)
+			new SelectEntry("Quit", Game::quit),
+			new BlankEntry()
```
Makkkkus, line 58: Why did you do this?

Makkkkus:
Very good changes!

lino-levan:
Any other questions?

Makkkkus:
No

src/main/java/minicraft/saveload/Load.java
```diff
@@ -399,6 +400,8 @@ private void loadPrefs(String filename) {
 			Game.input.setKey(map[0], map[1]);
 		}
 
+		Renderer.setAspectRatio(); // Sets the aspect ratio of the game window.
+
```
Makkkkus, line 404: This should be at line 381.

Makkkkus (approved):
Very good

Issue #257, opened 2022-01-13
Individual textures per sprite


AstroCat22:
That's about it.

lino-levan:
@Makkkkus thoughts?

Makkkkus:
It could be a long-term goal, yes

Makkkkus:
This probably won't happen. The current system is good enough for its use-case.

Makkkkus:
Will not implement because it would add many unnecessary disk operations, and be confusing to work with. Parts are already implemented, though, such as with skins.

PR #302, opened 2022-04-08
datadriven structures, recipes and books and new ids

KalmeMarq:
Structure DataDriven
```go
{
  key: {
    X: string // item id
  },
  data: [
    "XX X ",
    ...
  ],
  furniture: [
    {
      x: int,
      y: int,
      id: string // furniture id
    }  
  ]
}
```
Recipe DataDriven
```go
{
  for: ("workbench" | "craft" | "loom" | "anvil" | "enchant" | "oven" | "furnace")[]
  requires: [
    {
      item: string,
      amount?: int,
    }
  ],
  product: {
    item: string,
    amount?: int
  }
}
```
Book
```json
{
  "lines": [
    "localized.text.line1",
    "",
    "localized.text.line2"
  ]
}
```
Trying to add a more consistent id system.
Added new registries for items, tiles (and furniture) that uses a new id system. Start using "my_item" instead of "My Item"/"mY Item"/"My iTem"/etc.
If it can't find in the new registry it will use the old one.


BenChueng0422:
Tiles...why not use numeric ids?

KalmeMarq:
"You can use txt for book localizations..."
wdym? the book.json has the lines. each line is a localized text. and the translation is in the loc files.

"Tiles...why not use numeric ids"
bc how am I suppose to know the id of a tile. it's better to work with string ids.

"I have suggestion for structures, use SDT instead."
I'm not sure how std works.

Makkkkus:
I prefer what KalmeMarq has done here.

BenChueng0422:
For IDs I have an idea about that (I've sent it on Discord so I send the link here).
https://discord.com/channels/280723930942013440/280724207867002880/961884588244877312

KalmeMarq:
I might delete this pr until the std data and (id system has a consensus). I'll redo this after and add the loot tables together.

Makkkkus:
That's bad coding practice. We should only use string IDs, not numeric ones. It has to do with code readability.

Makkkkus:
Splitting books up into several lines seems kinda pointless to be honest.

BenChueng0422:
That makes sense, but we could use separate txt for other languages too.

KalmeMarq:
I think using the existing loc files is more convenient. U would have to create loc.count * 5 files when loc file could do just fine.

Litorom:
This is interesting and a big change to the system.
This also may be helpful in doing #257.

Litorom:
This may also make things easier for @PseudoDistant’s fabric system.

PR #307, opened 2022-04-20
Localization: More localization options and ES-ES finished

Litorom:


src/main/java/minicraft/item/TileItem.java
```diff
@@ -73,7 +73,7 @@ protected TileItem(String name, Sprite sprite, int count, String model, List<Str
 		this.model = model.toUpperCase();
 		this.validTiles = new ArrayList<>();
 		for (String tile: validTiles)
-			 this.validTiles.add(tile.toUpperCase());
```
Makkkkus, line 76: This breaks all of the notifications in the file, and possibly more. Remove all changes to this file and I will merge.

src/main/java/minicraft/item/TileItem.java
```diff
@@ -91,10 +91,10 @@ public boolean interactOn(Tile tile, Level level, int xt, int yt, Player player,
 		
 		String note = "";
 		if (model.contains("WALL")) {
-			note = Localization.getLocalized("minicraft.notification.invalid_placement") + " " + Tiles.getName(validTiles.get(0)) + "!";
+			note = Localization.getLocalized("minicraft.notification.invalid_placement") + " " + Localization.getLocalized(this.validTiles.get(0)) + "!";
 		}
 		else if (model.contains("DOOR")) {
-			note = Localization.getLocalized("minicraft.notification.invalid_placement") + " " + Tiles.getName(validTiles.get(0)) + "!";
+			note = Localization.getLocalized("minicraft.notification.invalid_placement") + " " + Localization.getLocalized(this.validTiles.get(0)) + "!";
 		}
```
Makkkkus, line 98: I did say delete all changes in the file; delete ALL the changes in this file.

Litorom: But it localizes the tile. Otherwise it will say the translated text with the english tile

Makkkkus: It won't be localized, since the key is wrong, no matter what you do. So just keep it as it is.

Litorom: Wdym the key is wrong? It translates it correctly

Makkkkus: Then don't translate it. We can fix that issue properly when we get to that point, but that is something I will have to do later, not something you will do now.

PR #284, opened 2022-02-28
Using JSON for advanced data storage

BenChueng0422:
This update only affects Level Data (extradata).
SDT is based on NBT.

Closes #294

PR #309, opened 2022-04-21
Using JSON for advanced data storage

### LINKED ISSUE #294: Using JSON for data storage of Items and Tiles, also for save files on 2.1.0
Game Update Limitation

The work is expected which consists of the all formatted files which are saved in the world save. Including `Entities.miniplussave`, `Game.miniplussave`, `Inventory.miniplussave`, `Level[0-5].miniplussave`, `Level[0-5]data.miniplussave` (will be renamed as `Level[0-5].miniplussavedata` temporally) and `Player.miniplussave`.

**Describe alternatives you've considered**

**Additional context**
Currently, I am working on the LevelData storing. I am deciding to work on Inventory and Level storages now.
Anymore working progresses will be shown here I guess.

BenChueng0422:
A reopen of #284
Closes #294

The JSON format of Level[0-5], Player, Inventory, Game have been finished.


BenChueng0422:
@Litorom I think it is because I have never committed any to the repository.

Makkkkus (commented):
I really like the rework of the save system. Just make sure it doesn't break any past saves.

src/main/java/minicraft/item/Item.java
```diff
@@ -15,16 +17,19 @@ public abstract class Item {
 	
 	private final String name;
 	public Sprite sprite;
+	public JSONObject data;
```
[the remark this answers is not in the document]
BenChueng0422: It wouldn't be a problem, each item should have its own data even data might be empty.

src/main/java/minicraft/level/Level.java
```diff
@@ -127,7 +131,7 @@ public Level(int w, int h, long seed, int level, Level parentLevel, boolean make
 		this.w = w;
 		this.h = h;
 		this.seed = seed;
-		short[][] maps; // Multidimensional array (an array within a array), used for the map
+		AbstractMap.SimpleEntry<short[], JSONObject[]> maps; // Multidimensional array (an array within a array), used for the map
```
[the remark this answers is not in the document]
BenChueng0422: each tile should have its initial data.

src/main/java/minicraft/item/Item.java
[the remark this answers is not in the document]
Makkkkus: Sure, go ahead

src/main/java/minicraft/level/Level.java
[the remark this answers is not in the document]
BenChueng0422: It also won't be a problem unless the save file is too large.

src/main/java/minicraft/saveload/Load.java
```diff
@@ -437,99 +456,198 @@ private void loadUnlocks(String filename) {
...
+						}
+	
+						if(l == World.minLevelDepth+1 && tilename.equalsIgnoreCase("LAPIS") && worldVer.compareTo(new Version("2.0.3-dev6")) < 0) {
+							if(Math.random() < 0.8) // don't replace *all* the lapis
+								tilename = "Gem Ore";
+						}
+						tiles[tileArrIdx] = Tiles.get(tilename).id;
+						JSONObject obj;
+						try {
+							obj = (JSONObject)Tiles.get(tilename).getClass().getMethod("getDefaultData").invoke(null, new Object[0]);
```
Makkkkus, line 520: Don't these calls have a negative performance impact?

src/main/java/minicraft/saveload/Load.java
```diff
@@ -437,99 +456,198 @@ private void loadUnlocks(String filename) {
...
+						parent.printLevelLoc("INCONSISTENT STAIRS detected; placing stairsDown", p.x, p.y);
+						parent.setTile(p.x, p.y, Tiles.get("Stairs Down"));
+					}
 				}
 			}
 		}
 	}
 	
-	public void loadPlayer(String filename, Player player) {
+	public void loadPlayer(String filename, Player player) throws JSONException, IOException {
```
Makkkkus, line 611: These should absolutely not throw exceptions. Errors should be handled where they occur, so you can write a proper error message.

Makkkkus (commented):
First, the only thing that I see resonable to implement into the current version are the tile id datatype changes. Second, this really conflicts with the changes I want to do with the save system, as the changes you have made only partly address my issues.

I don't think adding complexity to data stored by tiles, and especially items, is a good thing; which is the whole point of this pull request. I can accept the use of json data for tiles, as tiles may need more data in the future, but I still don't see why we need data instead of just adding variables to the classes, like it always has been.

The use of checks for safe execution and documentaion has also been very weak, especially with the lack of comments and error messages.

My alternative solution for saving and loading data is adding a class at the end of each file that handles json serialization in a safe and clean way. This is a fairly standard way of doing things, to my knowledge.

I'm also going to make it very clear that I am not targeting you, but that changes as big as this have extremely high standards for me to merge it. Changes like this are also very tedious for me to work with, especially when it is only me maintaining this, and there is certainly a lot I have missed. So, to be honest, I don't really want massive pull requests, but this isn't that bad compared to the data driven functionality pull request, which was a lot to take in.

src/main/java/minicraft/item/Item.java
```diff
@@ -15,16 +17,19 @@ public abstract class Item {
...
```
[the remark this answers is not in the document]
BenChueng0422: The item data would have durability and book content. (I forgot about that.)

src/main/java/minicraft/level/Structure.java
```diff
@@ -40,9 +43,16 @@ public void draw(Level level, int xt, int yt) {
 			 level.add(furniture.get(p).clone(), xt+p.x, yt+p.y, true);
 	}
 
-	public void draw(short[] map, int xt, int yt, int mapWidth) {
-		for (TilePoint p: tiles)
+	public void draw(short[] map, JSONObject[] data, int xt, int yt, int mapWidth) {
```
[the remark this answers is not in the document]
BenChueng0422: Since they need initial data, would you like initialize the data after the level is generated or initialize the data when they are generated?

src/main/java/minicraft/saveload/Load.java
BenChueng0422: I think this is the only way to get the initial data of the tile given since the initial data can only be given by static method...I think it shouldn't be a method instead hmm

src/main/java/minicraft/level/Structure.java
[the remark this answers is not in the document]
Makkkkus: I would rather want to initialize data when the class is loaded.

src/main/java/minicraft/core/Renderer.java
```diff
@@ -281,7 +281,7 @@ private static void renderGui() {
 		if (player.activeItem instanceof ToolItem) {
 			// Draws the text
 			ToolItem tool = (ToolItem) player.activeItem;
-			int dura = tool.dur * 100 / (tool.type.durability * (tool.level+1));
```
Makkkkus, line 284: See, I find things like these problematic. Since now you just removed the static variable in the class and replaced it with a json variable.

src/main/java/minicraft/level/Structure.java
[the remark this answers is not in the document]
BenChueng0422: That would need more step on tick(), to check if the properties are exist.

src/main/java/minicraft/level/tile/DoorTile.java
```diff
@@ -18,6 +18,10 @@ public class DoorTile extends Tile {
 	protected Material type;
 	private Sprite closedSprite;
 	private Sprite openSprite;
+	public final JSONObject initialDefaultData = new JSONObject();
+	{
+		initialDefaultData.put("closed", 0);
```
Makkkkus, line 23: In this exsample. There would be no problem to make closed a variable in the class.

src/main/java/minicraft/level/Structure.java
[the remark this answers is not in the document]
BenChueng0422: So would you like to initialize data at the last line of Level constructor?

src/main/java/minicraft/saveload/Load.java
```diff
@@ -88,13 +86,16 @@ public class Load {
...
-	public Load(String worldname, boolean loadGame) throws JSONException, IOException {
+	public Load(String worldname, boolean loadGame) {
 		try {
 			worldVer = new Version(new JSONObject(loadFromFile(location + "/saves/" + worldname + "/Game" + extension, true)).getString("version"));
 		} catch (JSONException e) {
 			loadFromFile(location + "/saves/" + worldname + "/Game" + extension);
 			if (data.get(0).contains(".")) worldVer = new Version(data.get(0));
 			if (worldVer == null) worldVer = new Version("1.8");
+		} catch (IOException e) {
+			Thread.getDefaultUncaughtExceptionHandler().uncaughtException(Thread.currentThread(), e);
```
Makkkkus, line 97: This is very good.

src/main/java/minicraft/core/Renderer.java
BenChueng0422: Hmm it is because it doesn't show error since it is dynamic. There won't be compiling error, runtime error instead, when there is error.

Makkkkus: That is not a good thing. It is better to get compiler errors than runtime errors.

BenChueng0422: So...for item, dur and book shouldn't be dynamic.

Makkkkus: None of them needs to be dynamic. Just the same features could be added by making it a part of the class, in addition to more stability.

src/main/java/minicraft/level/Structure.java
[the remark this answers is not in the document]
BenChueng0422: I think using setTile would be better for structure.draw? setTile initializes tile.

src/main/java/minicraft/core/Renderer.java
BenChueng0422: Recovered the changes on this.

PR #311, opened 2022-04-21
Achievements: Added Notifications

Litorom:


src/main/java/minicraft/screen/AchievementsDisplay.java
```diff
@@ -152,8 +152,11 @@ private static boolean setAchievement(String id, boolean unlocked, boolean save)
         Logger.debug("Updating data of achievement with id: {}.", id);
 
         // Add or subtract from score
-        if (unlocked)
+        if (unlocked) {
             achievementScore += a.score;
+            // Tells the player that they achieved an achievement.
+            Game.notifications.add(Localization.getLocalized(id) + " " + Localization.getLocalized("Achieved!"));
```
Makkkkus, line 158: Why did you splice two locs together instead of creating a single new one?

Litorom: Done

PR #314, opened 2022-04-22
Fix Interaction of active item

### LINKED ISSUE #176: Make mining less random.
Currently, if you try to mine an ore there is a chance for it to not damage the ore but reduce the durability of the pickaxe.
I think after 3 or some more misses there should be a guaranteed hit to make mining ores less infuriating and make the possibility to not be able to hit any ores forever (due to randomness) not a problem.

BenChueng0422:
This includes all the tile interaction (e.g. TreeTile, OreTile, RockTile, etc.).
Reason: The List of the entities of the tile includes particles. Particles need time to be disappeared; potential cooldown occurred.
Fixes #176

### DIFF
--- a/src/main/java/minicraft/entity/mob/Player.java
+++ b/src/main/java/minicraft/entity/mob/Player.java
@@ -3,7 +3,6 @@ package minicraft.entity.mob;
 import java.util.ArrayList;
 import java.util.HashMap;
 import java.util.List;
-import java.util.stream.Collectors;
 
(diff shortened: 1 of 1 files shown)


Makkkkus:
The changes are redundant. Filtering is already done by the method

BenChueng0422:
Right, now should be fine.

Makkkkus:
Very good

PR #331, opened 2022-04-29
Resolves #329

### LINKED ISSUE #329: Prevent achievements from being obtained in creative mode
**Describe the bug**
Achievements can be achieved in creative mode.

Side note: Achievement text overflows off screen if too long. Going down to the gem level will trigger this.

**To Reproduce**
1. Open a creative mode world
2. Attempt to achieve an achievement
3. Achievement is obtained

**Expected behavior**
It is not obtained.

**Screenshots**
Screen Shot 2022-04-29 at 8 41 28 AM

BenCheung0422:
This should Resolve #329

### DIFF
--- a/src/main/java/minicraft/core/World.java
+++ b/src/main/java/minicraft/core/World.java
@@ -13,7 +13,7 @@ import org.tinylog.Logger;
 
 public class World extends Game {
 	private World() {}
-	
+
 	public static final int[] idxToDepth = {-3, -2, -1, 0, 1, -4}; /// This is to map the level depths to each level's index in Game's levels array. This must ALWAYS be the same length as the levels array, of course.
 	public static final int minLevelDepth, maxLevelDepth;
 	static {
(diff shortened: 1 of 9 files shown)


lino-levan:
Don't reformat the files. It creates a bunch of noise that makes it hard to see what has been changed.

Makkkkus (changes requested):
Instead of adding an if statement every time we want to give the player an achievement, we should simply add an `if (Game.isMode("creative")) return;` to setAchievement().

I will also add that the current title for the pull request doesn't tell us anything on a first glance, so you don't know what the pull request is about before you check the linked issue. Issues should never be linked in titles; there is a reason why the number is blue in the description but not the title.

Keep the formatting changes though.

lino-levan:
> Instead of adding an if statement every time we want to give the player an achievement, we should simply add an `if (Game.isMode("creative")) return;` to setAchievement().

I'm not sure if this is the best way to do this? There are some achievements that are not achieved in-game but in the menu. Maybe it's a overloaded method where there's a boolean flag to check whether or not it's in creative?

Makkkkus:
In that case change it to `if (checkCreative && Game.isMode("creative")) return;` and add that argument to the function.

Litorom:
I have an idea of how to do this.

Litorom:
Made PR #332

PR #334, opened 2022-05-01
Fix UI problem when notification is too long

BenCheung0422:


src/main/java/minicraft/core/Renderer.java
```diff
@@ -215,68 +216,70 @@ private static void renderGui() {
...
 			}
-			
+
 			if (Updater.notetick > 120) { // Display time per notification.
 				notifications.remove(0);
 				Updater.notetick = 0;
 			}
-			
+			List<String> print = new ArrayList<>();
+			for (String n : notifications) for (String l : Font.getLines(n, Screen.w, Screen.h, 0)) print.add(l);
```
Makkkkus, line 248: Encapsulate the two for statements in { }

PR #337, opened 2022-05-05
Changed potion colors to make them visible

lino-levan:
Also removed unused imports since those seem to keep crawling into the files somehow.

Old:
Screen Shot 2022-05-05 at 8 59 56 AM

New:
Screen Shot 2022-05-05 at 8 59 19 AM

Also a small thing to note here is that the order of potion effects changes between saves/loads. Haven't looked into why, nor will I since that is incredibly low priority.

### DIFF
--- a/src/main/java/minicraft/entity/mob/Mob.java
+++ b/src/main/java/minicraft/entity/mob/Mob.java
@@ -1,7 +1,5 @@
 package minicraft.entity.mob;
 
-import java.util.List;
-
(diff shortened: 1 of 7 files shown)


lino-levan:
By the way, I just want to clarify that by "changed potion colors", I just scaled up the brightness of the colors. I did not change the actual hue or saturation.

lino-levan:
Ideally we re-write the potion hud because it's really ugly atm.

PR #338, opened 2022-05-06
Resolve texture pack loading and TNT Damaging problem

BenCheung0422:
Resolve the texture pack loading problem.
Resolve the TNT damaging problem (damage mobs with negative damages) problem.
Resolve when the spawners are being picked up, the sprites always are shown with default sprite.
Ability to save the dead status.
Adjust TNT damage in order to cause damage with nearly 50 (with what was written in ExplodedTile#tick).


Makkkkus:
One pull request for each issue

PR #340, opened 2022-05-06
Ability to save the dead status.

BenCheung0422:


src/main/java/minicraft/screen/PlayerDeathDisplay.java
```diff
@@ -32,7 +33,11 @@ public PlayerDeathDisplay() {
 				Game.setDisplay(null);
 			}));
 		}
-
+		
+		entries.add(new SelectEntry("Save and Quit", () -> {
```
Makkkkus, line 37: Add this to an else for the if statement above

PR #358, opened 2022-06-03
Fix of these ones before most of recipes display

pelletsstarPL:
added 2 spaces for better formatting.


Makkkkus:
This removes the name of the recipie.

TheBigEye:
Should be something like:
```java
 return super.toString().replace("1 ", "") + (recipe.getAmount() > 1 ? " x" + recipe.getAmount() : "");
```

PR #364, opened 2022-06-06
Stackable item limit

### LINKED ISSUE #305: Stackable item limit and inventory limit
Stackable items and inventories should have a limit to how many items they can contain. This should fix issues like #80.

A cap on how many items we can put in a stack, and a cap on how many items we can put in our inventory.

BenCheung0422:
Resolves #305


TheBigEye:
### [BUG] Items pickup, chests, and inventory...
- Chests behave weird:
    - There are times when the chest cursor cannot be moved
    - The items that are added from the chest are not shown in the player's inventory in the chest display, and neither does the inventory sound (that is, the player's inventory is not updated, when opening and closing the display they are already shown )
    - Cannot add the item searched by the searcher bar to the player's inventory and vice versa
    - Also, trying to add too many items from the chest to the player's inventory crashes the game (`java.lang.IndexOutOfBoundsException: Index 267 out of bounds for length 260`)
- Items dropped by breaking a tile cannot be picked up in creative mode.
- there are times that you can not collect the items dropped in survival mode

TheBigEye:
- The player can craft items in creative mode (by doing so they are not stored in the inventory, but are dropped when leaving the Crafting Display)
- When taking a Furniture from the inventory (Chest, Anvil, Workbench, etc) in creative mode, when placing it, and, when wanting to return it to the inventory, the item is dropped, maybe it is because of the limit of items in the inventory?
    - The same thing happens with normal items, like potions, ores, etc.
- In creative mode the player should be able to pick up items even if there is a limit on the inventory.
    - due to this limit, it is also not possible to empty a chest in creative mode

Litorom:
How will this work with old worlds? Will it keep the overflow items until taken out or will they be deleted, or dropped.

BenCheung0422:
@TheBigEye 
* To have minimum inventory use, all unnecessary usage are combined together.
* Emptying the chest in creative mode can be implemented with #360.
* Since the inventory size limit is 27, the crafted items are dropped if the inventory is full. This also work in getting back the held item to the inventory.

@Litorom It should now work with old worlds. A death chest is generated for the exceed items.

Issue #203, opened 2021-09-20
Quest system


Makkkkus:
**Describe the solution you'd like**
Create a quest system to make the game more RPG-like. This doesn't need to be used, but the solution would make modding the game even more interesting. This also allows for tutorials.

Litorom:
Seem relatively done at this point, reopen if necessary.

PR #383, opened 2022-07-14
Quest system part 2

BenCheung0422:
Implementation of #203 
The `tutorials` and `quests` settings implementation. Also recipe unlock implementation (in tutorials).
The quests are not completed. The quests will be completed in the later pull requests.


Makkkkus (commented):
I have a little feedback for you: I wish the `unlocks` entry in the quests.json file to be renamed to `leads_to`, while `recipies` is put into a new entry called `unlocks`. This is so we can make quests unlock more than just recipies, in the future.

I also wish for tutorial quests to be put in a separate file called `tutorials.json`, and allowing for more than one tutorial.

PR #387, opened 2022-07-20
Create CrashHandler for error handling

### LINKED ISSUE #386: A more clear error handler
The current error handle is not clear or handled at all enough.

We should make the error handling so as to make it more user-friendly.
For example a standard function to handle errors that affect the game system. And maybe a even more detailed error pane.

BenCheung0422:
Also applied to some exception catches.
Resolves #386

### DIFF
--- a/src/main/java/minicraft/core/CrashHandler.java
+++ b/src/main/java/minicraft/core/CrashHandler.java
@@ -1,0 +1,264 @@
+package minicraft.core;
+
+import minicraft.core.CrashHandler.ErrorInfo.ErrorType;
+import minicraft.core.io.ClipboardHandler;
+import minicraft.network.Analytics;
+
+import java.awt.*;
+import java.io.PrintWriter;
+import java.io.StringWriter;
+import java.util.concurrent.Future;
+
+import javax.swing.BorderFactory;
+import javax.swing.BoxLayout;
+import javax.swing.JButton;
+import javax.swing.JDialog;
+import javax.swing.JFrame;
+import javax.swing.JLabel;
+import javax.swing.JOptionPane;
+import javax.swing.JPanel;
+import javax.swing.JScrollPane;
+import javax.swing.JTextArea;
+import javax.swing.SwingConstants;
+import javax.swing.UIManager;
+import javax.swing.WindowConstants;
+
+import org.tinylog.Logger;
+
+import kong.unirest.Empty;
+import kong.unirest.HttpResponse;
+
+public class CrashHandler {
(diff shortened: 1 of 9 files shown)


src/main/java/minicraft/core/Renderer.java
```diff
@@ -66,20 +67,16 @@ public static SpriteSheet[] loadDefaultSpriteSheets() {
 			// These set the sprites to be used.
 			itemSheet = new SpriteSheet(ImageIO.read(Objects.requireNonNull(Game.class.getResourceAsStream("/resources/textures/items.png"))));
 			tileSheet = new SpriteSheet(ImageIO.read(Objects.requireNonNull(Game.class.getResourceAsStream("/resources/textures/tiles.png"))));
-			entitySheet = new SpriteSheet(ImageIO.read(Objects.requireNonNull(Game.class.getResourceAsStream("/resources/textures/entities.png"))));
+			entitySheet = new SpriteSheet(ImageIO.read(Objects.requireNonNull(Game.class.getResourceAsStream("/resurces/textures/entities.png"))));
```
Makkkkus, line 70: Ah, you forgot to fix this after testing.

src/main/java/minicraft/core/io/FileHandler.java
```diff
@@ -111,8 +112,8 @@ else if (ifExisting == RENAME_COPY) {
 				Path newFile = new File(newFilename).toPath();
 				try {
 					Files.copy(file, newFile, StandardCopyOption.REPLACE_EXISTING);
-				} catch(Exception ex) {
-					ex.printStackTrace();
+				} catch (Exception ex) {
+					CrashHandler.errorHandle(ex);
```
Makkkkus, line 116: Ideally, these exceptions would be handled automatically by the AWT.

Makkkkus (commented):
This is pretty good, but I notice some small issues here and there.

src/main/java/minicraft/core/Renderer.java
BenCheung0422: Ah.

src/main/java/minicraft/core/CrashHandler.java
```diff
@@ -0,0 +1,264 @@
...
+			public final int exitCode;
+			public final String message;
+
+			ErrorType(int exitCode, String message) {
+				this.exitCode = exitCode;
+				this.message = message;
+			}
+		}
+	}
+}
```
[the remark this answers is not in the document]
BenCheung0422: ok

Issue #313, opened 2022-04-22
ID and Resources rework


BenChueng0422:
**Is your feature request related to a problem? Please describe.**
Fix the compatibility and readability of IDs.

**Describe the solution you'd like**
Rework of the IDs, also the resources like structures.

Makkkkus:
Reliant upon #315

PR #391, opened 2022-07-30
Adds more localizations with IDs

BenCheung0422:
Added more localizations.
Remapped localizations with IDs except items, tiles and entities. (Partly implemented #313)
Some other small tweaks and modifications.
Added minicraft.screen.PopupDisplay.
Removed minicraft.screen.WorldEditDisplay.
Better Java 8 compatibility.
Added more tags to logging.
Added key binding conflict check.


src/main/java/minicraft/core/Game.java
```diff
@@ -68,22 +68,26 @@ public static void exitDisplay() {
...
+		// Applying Game#debug first.
+
 		Analytics.GameStartup.ping();
 
-		Initializer.parseArgs(args); // Parses the command line arguments
+		// Load default loc.
+		Localization.loadLanguage();
+		/* DO NOT trigger any other classes before this.
+		* Including static initialization.
+		*/
```
Makkkkus, line 80: Usually, we don't add comments under a statement.

Makkkkus (commented):
Couldn't really find anything wrong with this, so good job!

PR #393, opened 2022-08-01
Controller Support

### LINKED ISSUE #375: Controller Support
Controller Support is a highly recommended feature request made previously. It’s great feature to add for those who prefer controllers over keyboards/mouse.

Add controller support

**Additional context**
Support for the following:
- [ ] Xbox Controller

Litorom:
Controller Support:

Current Binds
- A = Attack/Select
- B = Back
- X = Inventory
- Y = Craft
- Menu/Start = Pause
- Joystick/DPad = Movement
- DPad (Menus) = Entry Movement

Tested Controllers:

- Xbox

(Binds may change especially with Quest additions and additional keybinds)
closes #375

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -19,14 +19,21 @@ mainClassName = 'minicraft.core.Game'
 
 repositories {
     mavenCentral()
+    maven { url "https://plugins.gradle.org/m2/" }
+    maven { url "https://jitpack.io" }
 }
 
 dependencies {
     implementation 'org.jetbrains:annotations:23.0.0'
-    implementation 'org.json:json:20211205'
-    implementation 'com.konghq:unirest-java:3.13.7'
+    implementation 'org.json:json:20220320'
+    implementation 'com.konghq:unirest-java:3.13.10'
     implementation 'org.tinylog:tinylog-api:2.4.1'
     implementation 'org.tinylog:tinylog-impl:2.4.1'
+    implementation 'com.badlogicgames.gdx:gdx:1.11.0'
+    implementation 'com.badlogicgames.gdx:gdx-box2d:1.11.0'
+    implementation 'com.badlogicgames.gdx:gdx-controllers:1.9.13'
+    implementation 'uk.co.electronstudio.sdl2gdx:sdl2gdx:1.0.4'
+    implementation 'com.github.WilliamAHartman:Jamepad:1.4.0'
 }
 
 java {
--- a/src/main/java/minicraft/core/io/ControllerHandler.java
+++ b/src/main/java/minicraft/core/io/ControllerHandler.java
@@ -44,6 +44,9 @@ public class ControllerHandler extends ControllerManager {
 		buttonmap.put("PICKUP", ControllerButton.LEFTBUMPER);
 
 		buttonmap.put("PAUSE", ControllerButton.START);
+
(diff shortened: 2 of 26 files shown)


src/main/java/minicraft/core/io/ControllerHandler.java
```diff
@@ -0,0 +1,30 @@
...
+	private ControllerIndex controllerIndex;
+
+	public ControllerHandler() {
+		controllerManager.initSDLGamepad();
+		controllerIndex = controllerManager.getControllerIndex(0);
+		controllerManager.update();
+		try {
+			Logger.debug("Controller Detected: " + controllerManager.getControllerIndex(0).getName());
+		} catch (ControllerUnpluggedException e) {
+			Logging.CONTROLLER.debug("No Controllers Detected, moving on.");
```
BenCheung0422, line 18: I think these two could also Logging.CONTROLLER

BenCheung0422 (changes requested):
Why we don't merge these two InputHandler and ControllerHandler into one InputHandler? Using the mapping instead of creating a new class for it. We need it to be able to maintained easier. We could use a small object to indicate the key mappings of both keyboard and controller, then that's all.

build.gradle
```diff
@@ -19,14 +19,21 @@ mainClassName = 'minicraft.core.Game'
...
     implementation 'org.jetbrains:annotations:23.0.0'
-    implementation 'org.json:json:20211205'
-    implementation 'com.konghq:unirest-java:3.13.7'
+    implementation 'org.json:json:20220320'
+    implementation 'com.konghq:unirest-java:3.13.10'
     implementation 'org.tinylog:tinylog-api:2.4.1'
     implementation 'org.tinylog:tinylog-impl:2.4.1'
+    implementation 'com.badlogicgames.gdx:gdx:1.11.0'
+    implementation 'com.badlogicgames.gdx:gdx-box2d:1.11.0'
+    implementation 'com.badlogicgames.gdx:gdx-controllers:1.9.13'
```
Makkkkus, line 34: Do we need these three when you have Jamepad installed?

Makkkkus (commented):
Couldn't find anything really bad. I agree with Ben's review, and encourage you too look into that. Other than that, this is very good!

Litorom:
> Why we don't merge these two InputHandler and ControllerHandler into one InputHandler? Using the mapping instead of creating a new class for it. We need it to be able to maintained easier. We could use a small object to indicate the key mappings of both keyboard and controller, then that's all. Also, the controller keys are not able to be changed in KeyInputDisplay.

For merging InputHandler and ControllerHandler, I attempted that, it was a bit problematic. Also in most games, controller binds are never changeable and are pre-defined

src/main/java/minicraft/core/io/ControllerHandler.java
Litorom: Fixed, I missed this.

build.gradle
Litorom: Yes, they are required for basic controller functions, with this jamepad and it's other dependency works fine.

BenCheung0422 (approved):

BenCheung0422:
As a placeholder of controller support before 3.0, it would be good enough.

PR #397, opened 2022-08-09
Fix Stream.toList() usage.

El-Virus:
Stream.toList() is implemented in java 16, we're targeting java 8. With this PR minicraft will be able to be built again.


Makkkkus:
This pull request is actually perfect. It fits all the requirenments for a good pull request, so I will highlight this.

PR #401, opened 2022-08-24
Update ChangeLog.txt

BenCheung0422:
Add 2.2.0 change log.


Makkkkus (changes requested):
"Obsidian knight" should not be capitalised.

Makkkkus (approved):

Litorom (changes requested):
Remove Controller Support, it’s a bit buggy. Doesn’t affect normal gameplay therefore it’s fine if it’s coded in and published.

BenCheung0422:
@Litorom I think maybe it still can be kept until the change of the system. As you said that it doesn't affect normal gameplay.

Litorom:
> @Litorom I think maybe it still can be kept until the change of the system. As you said that it doesn't affect normal gameplay.

No because if a user knows there is controller support when it isn’t finished.

Litorom (approved):
Okay, that’s better.

PR #405, opened 2022-10-07
if --savedir is present, use it as final gameDir

KalmeMarq:
Right now, when the --savedir is specified the game appends "/.playminicraft/mods/Minicraft Plus" to it. However, this should not be the case. If the user gives a path to use as the save dir then that should be the final path to use.

### DIFF
--- a/ChangeLog.txt
+++ b/ChangeLog.txt
@@ -16,7 +16,7 @@ Minicraft+ version 2.2.0
 * Changed a better logging
 * Fixed seed random problem
 * Improved the localization support
-* Changed how the game uses the --savedir command line argument. Now, when specified, the game will use it as the final game directory path instead of adding an extra '.playminicraft/mods/Minicraft_Plus' to it.
+* --savedir, if present, no longer is appended an extra '.playminicraft/mods/Minecraft Plus' to it
 
 Minicraft+ version 2.1.0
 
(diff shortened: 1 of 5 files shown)


Makkkkus (changes requested):
Add it to the changelog

Makkkkus (dismissed):

ChangeLog.txt
```diff
@@ -16,6 +16,7 @@ Minicraft+ version 2.2.0
 * Changed a better logging
 * Fixed seed random problem
 * Improved the localization support
+* Changed how the game uses the --savedir command line argument. Now, when specified, the game will use it as the final game directory path instead of adding an extra '.playminicraft/mods/Minicraft_Plus' to it.
```
Makkkkus, line 19: You should probably rewrite this to make it shorter and simpler like the entries above it.

src/main/java/minicraft/core/io/FileHandler.java
```diff
@@ -41,31 +43,41 @@ private FileHandler() {}
 	}
 
 
-	public static void determineGameDir(String saveDir) {
-		gameDir = saveDir + localGameDir;
-		Logging.GAMEHANDLER.debug("Determined gameDir: " + gameDir);
+	public static void determineGameDir(@Nullable String saveDir) {
```
Makkkkus, line 46: Perhaps add some javadoc for this function.

Makkkkus (changes requested):
Some small changes

KalmeMarq:
I don't know if the grammar is the best.

Makkkkus (approved):
All my problems are addressed. Good job!

ChristofferHolmesland (approved):
Looks good. Commit messages could be more clear.

PR #408, opened 2022-11-20
More detailed filename validation check

BenCheung0422:
This filename validation check checks more deeply with the corresponding filesystem and/or operation system.
This uses patterns to validate filename. If we use `File#createFile()`, that would affect the speed depends on disk I/O efficiency.


ChristofferHolmesland (approved):
Looks good :)

PR #410, opened 2022-11-22
Changes from ChangeLog.txt into markdown format

BenCheung0422:
As the semver format is used instead of the old versioning system, the structure can be described by markdown structure. As some modification can be described more clearly.


ChristofferHolmesland (approved):
I like this change

Makkkkus (approved):
Looks good

PR #413, opened 2022-11-30
Update ore tile textures

BenCheung0422:
Update the textures of the ore tiles depending on their item textures.
Also, trim the height requirement of skin images from 32 to 16 as the second row is not used.


Makkkkus:
I don't think this is neccecary. All of the changes seemingly only remove details from the textures.

Makkkkus (approved):

Makkkkus:
Add it to changelog maybe.

BenCheung0422:
Putting it to release changelog might be weird if you talking about the textures.
I will add the skin part.

Litorom (changes requested):
Please Update all relating items to the new color scheme. This includes tools, armor, weapons, and totem of air. Please also improve the shape of the ingots; they don’t look right.

ChristofferHolmesland (approved):
I like the changes :)

Makkkkus (approved):
Yeah, forgot about this. This is good.

PR #415, opened 2022-12-05
Fix a bug on world gen "trouble" popup and add more changelogs

BenCheung0422:
This is caused by a setting error.
https://discordapp.com/channels/280723930942013440/329379112684879882/1048647785043480697


Makkkkus (changes requested):
Create a changelog entry and I will accept. :)

Makkkkus (approved):

ChristofferHolmesland (approved):

PR #417, opened 2022-12-06
Resolve a feature about reacting toggles too fast

BenCheung0422:
Originally, when you press the toggle with a hit, it would response many times, which might not be intended.


Litorom (approved):
Okay

ChristofferHolmesland (approved):

PR #418, opened 2022-12-07
Reorganizing the quest system

### LINKED ISSUE #344: Recipies shouldn't be unlocked on game start
Currently, all recipies are unlocked upon the start of the game. This might be a lot to take in for a new player, and therefore we should limit it.

With the upcoming story update, we could tie the aquisition of new recipies to the quest system or similar. This will make the game story much more rewarding to play through.

**Additional context**
Partly requires #203 to function fully, but can function without.

BenCheung0422:
Reorganizing the quest system.
Resolves #344
Fixes rendering positioning problem of color code styled texts.
Separates tutorials from the entire quest system.
Recipes are unlocked only when one of the materials are obtained.


src/main/java/minicraft/item/Inventory.java
```diff
@@ -106,32 +107,47 @@ public int add(int slot, Item item) {
...
 			}
 		} else {
-			items.add(slot, item);
+			addItemToList(slot, item);
 			return 1;
 		}
 	}
 
+	private void addItemToList(Item item) { addItemToList(items.size(), item); }
+	private void addItemToList(int index, Item item) {
```
Makkkkus, line 137: Why do you need index here?

Makkkkus, line 137: These functions should be renamed so they explain _exactly_ what they are doing. This name does not tell us that it checks for quests.

Good that it is private.

Makkkkus (commented):
This is good. Haven't tested it so I can't approve it, but it looks good.

BenCheung0422:
Cool, but it is not ready and enough.

BenCheung0422:
I found that there might be a potential problem that the item handling on both breaking and interacting is not standardized. They are separated everywhere in almost every tiles.

BenCheung0422:
This pull request is now ready.

ChristofferHolmesland (approved):

Litorom (dismissed):
Skimmed through code looks good. Can’t fully personally test it at this time, however concept and code is good

ChristofferHolmesland (approved):

PR #403, opened 2022-09-04
Obsidian Knight (Redo)

Litorom:
Adds the second boss (OBK) of Minicraft replacing AWII

Closes #376 
Fixes #380, Fixes #407

PR #421, opened 2022-12-23
Simple fix Gradle run problem

BenCheung0422:
Mentioned by: https://github.com/MinicraftPlus/minicraft-plus-revived/pull/403#issuecomment-1363594648


Makkkkus:
Haven't we already made a file handler for this exact problem?

BenCheung0422:
I am not sure where you meant?

ChristofferHolmesland (approved):

ChristofferHolmesland (dismissed):

KalmeMarq:
There's actually another way of fixing this and also being able to remove the listResources methods with IDE and outside and other shenanigans.
Create a dummy file in the src/main/resources/, for example ".root". Get its URI and then try to get its Path and if it fails do FileSystems.newFileSystem(rootURI, Collections.emptyMap()) and then try getting the Path again. After you got the Path of the .root file get its parent Path. Now you can use the src/main/resources folder as a file system and get files, paths, etc. Works in both IDE and in production. Nicest solution.

TheBigEye (approved):

ChristofferHolmesland (approved):

src/main/java/minicraft/core/io/FileHandler.java
```diff
@@ -159,38 +165,28 @@ public FileVisitResult visitFileFailed(Path p, IOException ex) {
...
 
+		ArrayList<String> names = new ArrayList<>();
+		try (Stream<Path> paths = Files.walk(path)) {
+			Path finalPath = path;
+			paths.forEach(p -> names.add(finalPath.getParent().relativize(p).toString()));
 			return names;
 		} catch (IOException e) {
-			CrashHandler.errorHandle(e);
-			return names;
+			throw new RuntimeException(e); // FATAL ERROR
```
Makkkkus, line 189: A fatal error should probably stop the game. That's what fatal means.

Makkkkus (commented):
Looks good, but I can't approve yet.

src/main/java/minicraft/core/io/FileHandler.java
BenCheung0422: Currently, there is no crash report system, only popups are used. This error is thrown, unhandled error is caught by the handler.

ChristofferHolmesland (approved):

PR #423, opened 2022-12-26
Add glass bottle and made potion bottles reusable

### LINKED ISSUE #422: Please consider making glass flasks reusable

It feels almost like a bad joke when I find that the glass flasks are consumed when using potions, it's like the character eats up the whole thing. It kind of matters to playing experience since sand does not regenerate.  Also, old clothes disappear when new ones are put on, similar problem.

Add in the Recipe "Glass to Glass Flasks" to the Workbench, and change the Recipes in the Enchanter to be "Glass Flask & Material to Potion", and then when using potion let the flask remain. As for clothes, it would feel great if the previous ones simply appear in the inventory when new ones are put on.

**Additional context**
I'm not native English speaker, please forgive for possible grammar mistakes.

BenCheung0422:
Resolves #422 
Added glass bottle item; made empty potion into awkward potion; a new glass bottle is returned to inventory if potion is used.
The clothing item the player was wearing will be returned when a new one applied.

Changes compared to the issue:
- The recipe "Glass to Glass Flasks" is applied as `glass *3 -> glass bottle *1`.
- For the enchanter recipes, "Glass Flask & Material to Potion" is not applied at all. Instead, `potion` item is changed to `awkward potion` and it costs 1 `glass bottle` instead of `glass`. All potions are based on `awkward potion` items in the recipes.
- A `glass bottle` item is added to the inventory "directly" (the item is dropped on floor if the item cannot be added).
- A `clothes` item with the previous clothing color is added to the inventory "directly" when the new `clothes` item is applied.


Makkkkus (changes requested):
Not what was described in #423

Litorom:
The clothes should also be fixed as #422 talks about both.

Makkkkus (changes requested):
Again, this is not what was requested in the issue. Please actually do what was asked, or specifically describe that this doesn't resolve #423

BenCheung0422:
Alright, I edited the pull request comment.

ChristofferHolmesland (approved):

TheBigEye (approved):

Issue #400, opened 2022-08-15
Globol resource manager


BenCheung0422:
**Is your feature request related to a problem? Please describe.**
We need a standardization for IDs to prevent confusion and messively making multiple ID system.

**Describe the solution you'd like**
To have a standard ID system on *everything*, we need a resource manager. Tiles, items, entities can get the sprites; sound handler can get the sounds; static books can get the texts; localization handler can get the localization. They should all use the same ID to get the resources which are related to same target. For example: Minecraft uses `<namespace>:<path>` where path uses `/` and `.` for the separator and registry (`minecraft:block`/`minecraft:item`/`minecraft:entity_type`) also they separate IDs into that and translation IDs, we use `<path>` where path uses `.` for the separator. We need to consider our globol ID registry format.

PR #426, opened 2023-01-15
Start reorganizing the resource packs code

KalmeMarq:
I want to reorganize the Resource Packs code. Right now, everything is handled in the ResourcePackDisplay and that makes the file long, complicated and confusing.

I just wanted to make this first commit so I could create the pr for receiving feedback.


BenCheung0422:
This pull request includes the contents of #400, which is unverified by @Makkkkus and if it planned, it is considered to be done in [2.3](https://github.com/MinicraftPlus/minicraft-plus-revived/projects/6).

Makkkkus (commented):
Looks good.

Maybe the reloader classes should be static? They could also be merged into one class, since they all do the same thing.

BenCheung0422:
After my personal consideration, I think both #427 and this pull request (#426) could be done by 2.2.1, which means both of these two pull requests would be merged, as most of the changes are not making the same. However, there are still some minor parts that are duplicated in both pull requests, so the order (which is the dependant and which is the dependency) is still needed to be considered.

BenCheung0422 (commented):
I still think that both #427 and #426 (this) could be considered. Some of the changes are duplicated, and some have been done in my pull request, so some changes could be merged, and some could be organized.
I also suggest the compatible past pack version resolving, backward pack version support, like what I have done (not fully, though, but expected). I have already done the default (vanilla), directory, built-in classic, and zip support. World-bundled one is in progress. The idea of resource pack manager and separate reloading handlers are good, i.e., the organization, so I think they can be adjusted to fit those.
I think mine can be embedded into yours, but my pull request is quite big, so maybe I could just cherry-pick some commits from yours? However, that might not be the best practice since the mechanics of pull requests are not like that. 
If it is still only either mine or yours can be taken into account like you have said, I do not think that it is efficient as there will be more duplications.

src/main/java/minicraft/core/Game.java
```diff
@@ -59,6 +70,12 @@ public static void exitDisplay() {
 	static boolean running = true;
 	public static void quit() { running = false; }
 
+	public static final ReloadableResourceManager resourceManager = new ReloadableResourceManager();
+	public static ResourcePackManager resourcePackManager;
+	public static VanillaResourcePack defaultResourcePack;
+	public static boolean reloading;
```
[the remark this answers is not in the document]
BenCheung0422: The handler could just handle the displays by their corresponding classes. For the static classes, I think it is a big change for the game, so I think it is not the time for this.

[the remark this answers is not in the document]
BenCheung0422: Here, the display class for resource packs is `ResourcePackDisplay`, so it could handle the displays related to the resource pack system. About the display stack, if you mean the stacking of displays with handlers, it has already been implemented by an `ArrayDeque`.

BenCheung0422:
What about merging these two pull requests into one? Since there are many modifications in both of the pull requests targeting to improve the resource pack system.

BenCheung0422:
> Remove static classes and static variables[...]

These changes are not planned to do right now, but in long time goal for 2.3 and 2.4 updates while these changes made to the current resource pack system only applied to 2.2(.1) update. Although your statement is true, we still have to improve the system first before doing all this in long term goal.

BenCheung0422:
We could make suitable handlers without exceeded number of modifications to other codes.

BenCheung0422:
We are not willing to propose new changes based on future changes as they are not yet done or even started. So, now, these pull requests only do changes based on the current status of the codes, and that is the only way. The system still needs to be improved and more completed even before 2.3 and 2.4 updates. The current resource pack system is good enough for the first version, but it can still be updated and expanded for the next minor version.

src/main/java/minicraft/core/Game.java
```diff
@@ -59,6 +70,12 @@ public static void exitDisplay() {
 	static boolean running = true;
 	public static void quit() { running = false; }
 
+	public static final ReloadableResourceManager resourceManager = new ReloadableResourceManager();
+	public static ResourcePackManager resourcePackManager;
+	public static VanillaResourcePack defaultResourcePack;
```
BenCheung0422, line 75: I wonder if putting all these in Game class is good as it might make it more messy. I think they should be handled with a dedicated class.

src/main/java/minicraft/core/Game.java
```diff
@@ -77,6 +94,25 @@ public static void main(String[] args) {
...
+		// resourceManager.registerReloader(new TextureReloader());
+		resourcePackManager.setEnabledPacks(Arrays.asList("vanilla"));
+		Game.reloading = true;
+		Game.currentLoader = resourceManager.reload(resourcePackManager.getEnabledPacks(), () -> {
+			Game.reloading = false;
+			Game.currentLoader = null;
+		});
+
+		// TODO: remove reloading from this method so that the reload overlay can be shown >:(
+		// So remove the splash window too because is not needed. When reloading, no screen and level should be render.
```
BenCheung0422, line 115: I do not agree with this as the splash screen directly takes the image file from the jar, so it should not be affected.

src/main/java/minicraft/util/resource/reloader/SyncResourceReloader.java
```diff
@@ -0,0 +1,15 @@
+package minicraft.util.resource.reloader;
```
BenCheung0422, line 1: I think all the implementations of this class could be managed in this file.

src/main/java/minicraft/core/Renderer.java
```diff
@@ -114,16 +114,28 @@ public static void initScreen() {
...
+			if (readyToRenderGameplay) {
+				renderLevel();
+				if (player.renderGUI) renderGui();
+			}
 
+			if (display != null) // Renders menu, if present.
+				display.render(screen);
+
+			if (!canvas.hasFocus()) renderFocusNagger(); // Calls the renderFocusNagger() method, which creates the "Click to Focus" message.
+		}
```
BenCheung0422, line 138: I think this should really be handled in another display class, also referencing my previous comment to the reload field.

src/main/java/minicraft/core/Game.java
BenCheung0422: When the splash screen is shown, assets are being loaded. That should be better.

src/main/java/minicraft/core/Game.java
BenCheung0422: These can be integrated into a resource manager.

src/main/java/minicraft/util/resource/reloader/SyncResourceReloader.java
BenCheung0422: What if we want to maximally resolve previous version resource packs? Probably separating them is a better idea.

BenCheung0422:
If we are going to merge these 2 pull requests, some duplications should be resolved first. However, this first needs both me and you to agree with this.

BenCheung0422:
It may not delay the update as I would plan it to 2.2.1 patch. About the branches, I think it might not cause major problems.

Makkkkus:
Closing this because of lacking activity.

PR #429, opened 2023-01-17
Language fallback

BenCheung0422:
Added a default language fallback: if the localization is not localized under the selected locale, the localization under the default language (en-US) is used.


Makkkkus (changes requested):
The current functionality is intended. It is meant to show where you need to actually localise.

It is really bad to mix two languages together.

BenCheung0422:
Non-localized strings are always not friendly to normal players, especially localization keys; fallback localization is always needed, not all languages always have these localized.

BenCheung0422:
As showing where you need to actually localise has already been done by `unlocalized.txt`, fallback languages can not affect this ability. The debug locale I added could show where the displaying texts are not translated into translation IDs, I would suggest keeping it until the ID rework.

Makkkkus (dismissed):

Makkkkus:
Yeah, good point. I personally prefer having just the ids show up instead of making english show up. I don't like mixing languages.
I don't know what others think about this.

ChristofferHolmesland:
From a players perspective it's still playable if English is used instead of the IDs. As long as most of it is translated it should be fine to show English for the last parts.

Can we add a feature that shows the player if anything is missing when they select a language? For example red text saying "Warning: 2% of the text is not translated". Possibly a option to see exactly which parts are missing too, to make it easier to fix it.

BenCheung0422:
And what about a separated display showing all available languages instead of an entry?

BenCheung0422:
My opinion of IDs, IDs are not explaining the features and functionalities in details. The player would not be understandable only by the IDs instead of the detailed localized texts.

BenCheung0422:
As the external localization platform is used, the warnings and debugging unlocalization keys are no longer needed.

src/main/java/minicraft/core/io/Localization.java
```diff
@@ -1,28 +1,43 @@
...
+	private static final Set<String> DEFAULT_LOCALIZATION_OF_DEFAULT_LOCALE;
+
+	static {
+		try {
+			DEFAULT_LOCALIZATION_OF_DEFAULT_LOCALE =
+				new JSONObject(String.join("", Load.loadFile("/assets/localization/en-us.json"))).keySet();
+		} catch (IOException e) {
+			throw new RuntimeException(e); // CRITICAL ERROR (GAME ASSETS)
+		}
+	}
```
Makkkkus, line 40: It's probably better to put this in a constructor. I think we are going to need to move away from static variables to make the code easier to understand.

src/main/java/minicraft/core/io/Localization.java
```diff
@@ -1,28 +1,43 @@
...
+import java.util.stream.Collectors;
+
 public class Localization {
 
 	public static final Locale DEFAULT_LOCALE = Locale.US;
+	public static final Locale DEBUG_LOCALE = Locale.ROOT; // This locale is used for debugging;
 
 	private static final HashMap<Locale, HashSet<String>> knownUnlocalizedStrings = new HashMap<>();
 	private static final HashMap<String, String> localization = new HashMap<>();
+	private static final HashMap<String, String> fallbackLocalization = new HashMap<>();
```
Makkkkus, line 26: Merge localization and fallbackLocalization, and try to remove knownUnlocalizedStrings.

src/main/java/minicraft/screen/LanguageSettingsDisplay.java
```diff
@@ -0,0 +1,79 @@
...
+	@Override
+	public void render(Screen screen) {
+		super.render(screen);
+
+		ArrayList<String> list = new ArrayList<>();
+		Collections.addAll(list, Font.getLines(Localization.getLocalized("minicraft.displays.language_settings.display.accuracy"), Screen.w, Screen.h, 0));
+		for (int i = 0; i < list.size(); i++)
+			Font.drawCentered(list.get(i), screen, Screen.h - 8 * (list.size() - i), Color.GRAY);
+	}
+}
```
Makkkkus, line 79: Not a single line in this class has documentation. How is anyone supposed to understand on first sight what line 74-78 does? Or 33-44? Or 23-29?

src/main/resources/assets/localization/en-us.json
```diff
@@ -122,6 +124,8 @@
   "minicraft.displays.key_input.popup_display.confirm_reset": "Are you sure you want to reset all key bindings to the default keys?",
   "minicraft.displays.key_input.popup_display.press_key_sequence": "Press the desired key sequence",
   "minicraft.displays.key_input.title": "Controls",
+  "minicraft.displays.language_settings.title": "Language...",
+  "minicraft.displays.language_settings.display.accuracy": "(Language translations may not be 100%% accurate)",
```
Makkkkus, line 128: 100%

src/main/java/minicraft/core/io/Localization.java
```diff
@@ -110,12 +144,23 @@ public static void loadLanguage() {
 				localization.put(key, json.getString(key));
 			}
 		}
+
+		for (String text : unloadedLocalization.get(DEFAULT_LOCALE)) {
+			json = new JSONObject(text); // This JSON has been verified before.
+			// Put all loc strings in a key-value set.
+			for (String key : json.keySet()) {
+				fallbackLocalization.put(key, json.getString(key));
+			}
```
Makkkkus, line 153: These comments doesn't say anything more to the user other than what they can immediately see.

Makkkkus (changes requested):

src/main/java/minicraft/core/io/Localization.java
BenCheung0422: That means we would need to move static fields into object fields for all the classes.

Makkkkus: Yeah, but constants generally don't need to be changed. So only the static fields.

Makkkkus: Also noticed that the name of this field is wrong. It it should be camelCase.

src/main/resources/assets/localization/en-us.json
BenCheung0422: Alright.

src/main/java/minicraft/core/io/Localization.java
BenCheung0422: What?

src/main/java/minicraft/core/io/Localization.java
BenCheung0422: Public static fields are all constants, so they should be UPPER_SNAKE_CASE.

src/main/java/minicraft/core/io/Localization.java
```diff
@@ -145,11 +118,14 @@ public static void loadLanguage() {
 			}
 		}
 
-		for (String text : unloadedLocalization.get(DEFAULT_LOCALE)) {
+		// Language fallback
+		if (!selectedLocale.equals(DEFAULT_LOCALE)) for (String text : unloadedLocalization.get(DEFAULT_LOCALE)) {
```
Makkkkus, line 122: If and for mustn't be on the same line. And both should have brackets.

src/main/java/minicraft/core/io/Localization.java
Makkkkus: The comment's aren't very informative.

src/main/java/minicraft/core/io/Localization.java
Makkkkus: No, that set is not a constant.

BenCheung0422: Right, and that part you commented is removed now.

src/main/java/minicraft/core/io/Localization.java
Makkkkus: There's still not a comment that tells why this for loop exists.

BenCheung0422: I have added.

Makkkkus: It's kinda hidden in plain sight. Ideally code shouldn't need documentation, but in reality almost all code needs documentation. So, you need documentation, but since there are so many other lines that tell us absolutely no useful information, the actually useful comments are harder to find.

Be more picky about what you're actually documenting. Stop adding useless comments like: `Put all loc strings in a key-value set.` `Debugging locale.` `This JSON has been verified before.` (what does this even mean?) `Incrementing index.` `Decrementing index.`

BenCheung0422: That for loop snap is outdated. So, the comments you mentioned should be removed?

Makkkkus: Yes, but use your own common sense and decide what is relevant or not.

Makkkkus (dismissed):

Makkkkus (approved):

ChristofferHolmesland (dismissed):

Makkkkus (approved):

ChristofferHolmesland (dismissed):

PR #430, opened 2023-01-18
Better player's tile indicator

TheBigEye:
### Changes:
* Now corresponds to the size and outline of the tile (only the base tile, this doesn't fully fit tiles like trees or flowers)
* Now it will not render if the player is swimming (this is because it is uncomfortable as well as useless)


ChristofferHolmesland (approved):

BenCheung0422 (approved):
That's good, it is able to see the tile more clearly.

PR #431, opened 2023-01-18
Fix fishing data wrong path

TheBigEye:
### Context:
- When they made changes to the game's resource directory, they forgot to also update the path where the fishing system loot is obtained, causing the game to crash when trying to fish something


ChristofferHolmesland (approved):

BenCheung0422 (approved):
Ah, I missed this.

PR #439, opened 2023-02-14
Renaming from Item#clone to Item#copy

BenCheung0422:
`Object#clone()` is originally reserved by Java for the use for `Cloneable`, but in the method `Item#clone()` in the code, the way Java originally expected is not met. Therefore, renaming the method into `#copy()` would be better in practice.


ChristofferHolmesland (dismissed):

Makkkkus (approved):

ChristofferHolmesland (approved):

Litorom (approved):
Not null is a nice way to avoid certain buggy scenarios. Also copy sounds better named than clone. Clone may be a Chris J. thing.

ChristofferHolmesland (dismissed):

Makkkkus (approved):

PR #440, opened 2023-02-20
Making sheep and cow eating grasses

### LINKED ISSUE #201: More side activities separate from the storyline.
More activities to do apart from doing the main storyline. Ex: What to do after the second AirWizard is defeated.

Suggestions: Addition of End Game items and equipment. Ex: The second AirWizard will drop materials to craft a new pickaxe. This new Pickaxe can mine Cloud Ore and thus giving Cloud Ore a use.

### LINKED ISSUE #225: [FEATURE] PassiveMobs should eat certain food items dropped on the ground
There currently isn't any gameplay reason passive animals exist outside of food / wool. This is compounded because of issue #148.

Pigs should got to and eat potatoes. Cows/Sheep should go to and eat wheat.

BenCheung0422:
Resolves #225
Sheep and cow now eat grass on the floor. (grazing)
Tile interaction is added instead of interacting with item entities because making passive mobs eating dropped food items on the ground does not make sense. First, the item entities disappear in very short time. Second, items are dropped *only when the player drops items manually*, so, the chance of making this happening is too small. Third, meeting passive mobs is already a rare thing. These factors are the points needed to be considered into account.

### DIFF
--- a/src/main/java/minicraft/entity/mob/Sheep.java
+++ b/src/main/java/minicraft/entity/mob/Sheep.java
@@ -18,10 +18,7 @@ public class Sheep extends PassiveMob {
 	private static final LinkedSprite[][] sprites = Mob.compileMobSpriteAnimations(0, 0, "sheep");
 	private static final LinkedSprite[][] cutSprites = Mob.compileMobSpriteAnimations(0, 2, "sheep");
 
-	private static final int WOOL_GROW_TIME = 3 * 60 * Updater.normSpeed; // Three minutes
-
 	public boolean cut = false;
-	private int ageWhenCut = 0;
 
 	/**
 	 * Creates a sheep entity.
@@ -65,7 +62,6 @@ public class Sheep extends PassiveMob {
 			if (((ToolItem) item).type == ToolType.Shears) {
 				cut = true;
 				dropItem(1, 3, Items.get("Wool"));
-				ageWhenCut = age;
 				((ToolItem) item).payDurability();
 				return true;
 			}
--- a/src/main/java/minicraft/entity/mob/Sheep.java
+++ b/src/main/java/minicraft/entity/mob/Sheep.java
@@ -1,7 +1,5 @@
 package minicraft.entity.mob;
 
-import org.jetbrains.annotations.Nullable;
-
 import minicraft.core.Updater;
 import minicraft.core.io.Settings;
 import minicraft.entity.Direction;
@@ -11,6 +9,10 @@ import minicraft.item.Item;
 import minicraft.item.Items;
 import minicraft.item.ToolItem;
 import minicraft.item.ToolType;
+import minicraft.level.tile.GrassTile;
+import minicraft.level.tile.Tile;
+import minicraft.level.tile.Tiles;
+import org.jetbrains.annotations.Nullable;
 
 public class Sheep extends PassiveMob {
 	private static final LinkedSprite[][] sprites = Mob.compileMobSpriteAnimations(0, 0, "sheep");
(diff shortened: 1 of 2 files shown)


src/main/java/minicraft/entity/mob/Sheep.java
```diff
@@ -46,7 +45,14 @@ public void render(Screen screen) {
 	@Override
 	public void tick() {
 		super.tick();
-		if (age - ageWhenCut > WOOL_GROW_TIME) cut = false;
+		if (random.nextInt(1000) == 0) { // Grazing
```
Makkkkus, line 48: I think that it is better to have the random check inside the `tile instanceof GrassTile` if-statement. This makes it so that the sheep has to find a grass tile first, and then check if it wants to eat it.

src/main/java/minicraft/entity/mob/Sheep.java
```diff
@@ -46,7 +45,14 @@ public void render(Screen screen) {
...
+			Tile tile = level.getTile(x >> 4, y >> 4);
+			// If tall grasses are present, these are consumed and then turn into grass tiles.
+			if (tile instanceof GrassTile) {
```
Makkkkus, line 51: Shouldn't sheep eat tall grass? Sheep can't eat grass that is only a couple of centimeters long.

You can use classes in the other pull request, and write that that PR is required before this one.

Makkkkus (changes requested):

src/main/java/minicraft/entity/mob/Sheep.java
BenCheung0422: This is just a game, so we could just assert that sheep could eat any height of grass.

Makkkkus: Yes, but what is the point of having tall grass then?

BenCheung0422: First, we can have more fancy visual effect on plane grass tiles. Second, the sources and chances of getting wheat seeds and grass seeds are separated as dedicated sources.

src/main/java/minicraft/entity/mob/Sheep.java
BenCheung0422: I think it would be just fine. I would design it as when it wants to graze, first checking if there is any valid interactable tile under it, grazing it if so.

src/main/java/minicraft/entity/mob/Sheep.java
Makkkkus: I don't understand what is so problematic with making sheep eat tall grass. The comment even says it does such.

src/main/java/minicraft/entity/mob/Sheep.java
Makkkkus: First, you have to execute many more calls to random that you have to. Generating pseudorandom is more expensive than checking for a type. Secondly this makes it so that it practically never grazes, because the sheep has to be on a tile that it is able to graze on, which isn't certain.

src/main/java/minicraft/entity/mob/Sheep.java
BenCheung0422: Sheep would eat both tall grass and grass tiles.

Makkkkus: I'll accept that.

Makkkkus (approved):

Litorom (approved):
Implementation is good, sheep regrowing wool by eating grass tile is a nice concept. Even though I the concept seems copied.

PR #443, opened 2023-03-04
Optimize all imports

BenCheung0422:
As the imports and import orders are not standardized in all the java files, this should standardize all the imports acting as the same patterns.


Makkkkus (dismissed):

Makkkkus (approved):

ChristofferHolmesland (approved):

Litorom (approved):
I personally don’t know how Java handled imports but I’m assuming because it calls all imports of a class using ‘*’, doing used import rather than all saves data. So I approve.

ChristofferHolmesland (approved):

PR #445, opened 2023-03-04
Update color system and sheep coloring

BenCheung0422:
*This depends on #437 due to a fix of stackable item interaction on entities*

The current color system is pretty lack in customization and content, so this pull request complements and updates these parts.
- As sheep only has white wools, colored sheep is now available in natural spawn.
- There is now small chance in spawning colored sheep.
- Sheep and beds are now colorable.
- Dyes are added.
- 16 colors are implemented in dyes, wools, and colorable items and furniture.
- Adds a new recipe - `string * 3 -> white wool * 1`
- Adds 19 new flowers replacing original flowers.
- Dye Vat is added.
- Adds a new recipe - `stone * 10 + iron * 5 -> dye vat * 1`
- Old flower items and tiles are resolved as oxeye daisy.
- Watering can also generate newly added flowers.

When we have biomes supported, different biomes can have different ranges of flower variants generated.

(Discord thread: https://discordapp.com/channels/280723930942013440/1083334895763456072)


Litorom:
Seems incomplete, I don’t think we have all the wool colors shown.

BenCheung0422:
The coloring system has not been not discussed with a valid result (thread: https://discordapp.com/channels/280723930942013440/1083334895763456072). I need a decision about what colors should be implemented.

Litorom:
Maybe add dyeable beds and other color related items

BenCheung0422:
Should we add dying in clothing, more than the current ones? There are several ways we could accomplish this:
- Hard-coded clothing - 16 colors of clothing could be made by crafting in workbench like wools.
- Dyeable clothing plan A - cauldron is added to make dyed water, which could dye colors on clothing interacted by player.
- Dyeable clothing plan B - a dedicated menu is added to have ability to dye clothing.
With plans on dyeable clothing, the possible number of colors could be greatly greater (the exact numbers are still being calculated) than 16 (hard-coded) colored clothes.

BenCheung0422:
The clothing update will move to another pull request, so this pull request is now ready.

Makkkkus:
This should really be added to the changelog.

Makkkkus:
Why are all the changes in #437 included in this?

BenCheung0422:
It is difficult to include only parts of the branch as commits depend on each other.

Makkkkus:
Make it a draft and only include the actual changes that are different from the farming overhaul PR. Then say that it depends upon it. Merging this will be such a big mess that if you don't do it we will need to close this.

BenCheung0422:
The build would be failed.

Makkkkus:
It would fail until we merge the changes from #437, which when merged would make this compile, making this able to be merged.

BenCheung0422:
If I only includes changes apart from the farming branch, very complex rebase is required, but I think maybe there would be unideal changes made.

BenCheung0422:
This pull request entirely depends on #437.

src/client/java/minicraft/level/tile/Tiles.java
```diff
@@ -27,7 +27,27 @@ public static void initTileList() {
...
+		tiles.put((short)72, new FlowerTile("Orange Tulip", "orange_tulip"));
+		tiles.put((short)73, new FlowerTile("Pink Tulip", "pink_tulip"));
+		tiles.put((short)74, new FlowerTile("Red Tulip", "red_tulip"));
+		tiles.put((short)75, new FlowerTile("White Tulip", "white_tulip"));
+		tiles.put((short)76, new FlowerTile("Peony", "peony"));
+		tiles.put((short)77, new FlowerTile("Periwinkle", "periwinkle"));
+		tiles.put((short)78, new FlowerTile("Pink Lily", "pink_lily"));
+		tiles.put((short)79, new FlowerTile("White Lily", "white_lily"));
+		tiles.put((short)80, new FlowerTile("Poppy", "poppy"));
+		tiles.put((short)81, new FlowerTile("Violet", "violet"));
```
Makkkkus, line 49: As I have said before. Don't create new tiles -- make these variants of the same tile.

I have never given persmission to create new tiles. Especially for something as small as this.

Makkkkus (changes requested):

src/client/java/minicraft/level/tile/Tiles.java
Makkkkus: You still haven't fixed this. You can't resolve it before fixing it. I havew already said you don't have the permission to add any new tiles.

BenCheung0422: It would not make any sense if only rose tiles drop rose items and flower tiles of other or all flower models/varieties drop the same "flower" items because they are made into dyes. I would rather make all the flower tiles the same and make the game more monotonous instead. Is this what you want?

Makkkkus: Change the texture of the item to make it look like it could be any flower. You're making a problem out of nothing.

BenCheung0422: Then the placement of the item would not be as same as the original tiles and the recipes of dyes from flowers would not work anymore.

BenCheung0422:
This pull request should now be ready. I am not sure if we should implement dyeable glasses.

src/client/resources/assets/textures/item/bed.png
Litorom, line 1: We will not have this Minecraft-like item texture. Please revert to previous style.

Litorom (changes requested):
Please revert the bed style to previous for the item icon, it’s too blocky and doesn’t match the game. Then let players test it, when all is good I’ll merge it.

Makkkkus:
Looks like something weird is going on with the line endings. Perhaps check if they are commited as LF?

src/client/resources/assets/textures/item/bed.png
BenCheung0422: This kind of icon looks smallish and rigid. Actually the icon in Minecraft is a block render and angled. How do you define "Minecraft-like"? It actually looks more broken if it is not pixel- and direction-oriented as it is only 8x8.

src/client/java/minicraft/level/tile/Tiles.java
Makkkkus: Why can't you just make things simple and let it drop a random flower when you break it? The game doesn't need any more shit cluttering up the screen.

BenCheung0422: Then flowers would not be placeable and used as tile decoration.

Litorom: I would suggest having a flower tile that drops a random flower and allowing the placement of said flower with the variants, as a compromise.

Makkkkus: > Then flowers would not be placeable and used as tile decoration.

Yes, it can. Just make them TileItems. One item, one tile. Different flower variants for decoration.

Makkkkus: > I would suggest having a flower tile that drops a random flower and allowing the placement of said flower with the variants, as a compromise.

This is what I proposed.

BenCheung0422: Then there will be 2 tiles. One for flower, one for generation, kind of unnecessary.

BenCheung0422: Just one kind of flower tile throughout the whole game would be enough. No need for "random" variation on the same tile.

Makkkkus: > Then there will be 2 tiles. One for flower, one for random flower, kind of unnecessary.

No.

BenCheung0422: You have messed up here. I will combine the items into just variations of flower item. And as a result, there will be only 1 item and 1 tile for flowers, but they will be handled as variations by using item and tile data.

Litorom:
Can you fix the sprites to match the current sprites and no please artistic style changes.

BenCheung0422:
Besides your thoughts, I think several changes are quite reasonable.

- Change to texture of bed items
texture change for bed items
As several color variants are being added, it is essential to make them easier to be distinguished from each other even in item entity state. Those are small when dropped, so it might not be clear to know which color of the bed is. Originally there are only 6 pixels (even not rectangle) of (changeable) colors in texture, but in the new change, a larger block of 4\*6 colors pixels are used, which can be reasonably identified among others.
- Change to texture of bed entities
texture change for bed entities
I would not say that the artworks is really good (border color is taken from plank), but enough to reflect the details of the color variant pixels.
- Change to texture of wool items
texture change for wool items
This change relates to the elimination of unnecessary or unconformable black/dark pixels in texture. The excessive black pixels made the alike coloring tone (dull and non-color vibrant) of the texture, which should not be made to coloring-oriented elements.
- Change to texture of wool tiles
texture change for wool tiles
Style of shape is made as the change, also making them more distinguishable from other tiles. Please have a notice to the texture style of some other tiles:
classic dirt like tile textures
From left to right, up to down, they are dirt (surface), dirt (underground), dirt (dungeon), rock, grass and hard rock.
They all have the same shapes just with different colors, and thus similar textures, which cannot make them really identifiable among tiles. The similar can be applied to the original wool tiles. By the way, I believe that underground dirt is made as similar as rock because it might be tended to be like the ground of rocks underground. I think these textures might also have to be updated so as to have different looks.

Please note that we are improving and maintaining the game, but not necessarily just barely keeping all its elements blindly. If you have further suggestions, you are welcome.

Litorom:
The wool textures are fine, besides that I will not merge this until you follow the rules of design.

BenCheung0422:
Rules of design? Are you also talking about that 6 pixels of difference among color variants?

Litorom:
No, I’m asking you to follow the rules of the design of textures, currently you been messing with it. This WILL NOT be merged until you follow those rules. Do what I asked, in a real game company you would be in trouble to mess with the artwork that doesn’t match the aesthetic.

BenCheung0422:
I see, I first hear of the term "rules of design" of textures here.

Litorom:
Not quite what I asked for but it’s definitely an improvement. I’ll leave it at that, I’ll be testing out the PR now.

Litorom:
Fix the conflict and I’ll test it more

Litorom (changes requested):
if (worldVer.compareTo(new Version("2.2.0")) <= 0 && tileName.equalsIgnoreCase("FLOWER")) {
 				data[idx] = 0;
 			} else {
 				data[idx] = Short.parseShort(tileData);
 			}
 			
Does this just remove flower item/tile, if so it shouldn’t just remove it but replace it with a default variant. If rose change to rose, if flower tile change to a default flower tile. Don’t just remove it.

BenCheung0422:
Here it just reset the tile data for flower tile (and yeah, the version should be updated), with 0, something like default, which is Oxeye Daisy.

BenCheung0422:
Also note that originally rose was not placeable (just stackable item) and only "flower" item is placeable (as random flower tile).

PR #449, opened 2023-03-09
Clean up Level.java

BenCheung0422:
- Removes the redundant `Level#sparks` set for sparks as there are no difference if they are combined into `Level#entities` set.
- Fixes the potential problem about y-axis (part of coordinate setting) is not looped due to the always-false conditional expression in some furniture generation.
- Makes `#toArray()` function always use empty arrays.

### DIFF
--- a/src/main/java/minicraft/level/Level.java
+++ b/src/main/java/minicraft/level/Level.java
@@ -42,6 +42,7 @@ import java.util.List;
 import java.util.Random;
 import java.util.Set;
 import java.util.function.Predicate;
+import java.util.function.ToIntFunction;
 
 public class Level {
 	private final Random random;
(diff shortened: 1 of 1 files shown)


Makkkkus (commented):
Wasn't the sparks set added to prevent the sparks from flooding the entities set, and thus taking all the available entity slots?

BenCheung0422:
I don't that that should be taken into account. The set has already been dynamic.

Makkkkus:
Are you sure that this doesn't make the game lag when having many sparks?

BenCheung0422:
The time and memory taken is similar.

Makkkkus:
> The time and memory taken is nearly the same.

Are you sure that this doesn't make the game lag *when having many sparks*?

BenCheung0422:
How would you handle that then? The difference here between `entities` and `sparks` is that there are few more `#contains` and `#add` invokes, but as the collection is a set, the time complexity is O(1), which means the change on this is nearly none.

Makkkkus (approved):

ChristofferHolmesland (approved):

PR #451, opened 2023-03-10
Update despawn conditions

BenCheung0422:
- Updates the natural despawn conditions. (Old condition is totally replaced.)
  - Mobs farer than 40 tiles despawn randomly with 1/800 chance each tick after continuously 30 seconds away from players.
  - Mobs farer than 80 tiles despawn immediately.
- Updates the natural spawn conditions. (Old condition is totally replaced.)
  - Mobs randomly spawn at places between 10 and 40 tiles far from players. (30 trials for each player; distance check only available in 1 player at once.)
- Adds conditions to prevent despawning. (to-do; lack of ideas)

### DIFF
--- a/src/main/java/minicraft/entity/mob/Mob.java
+++ b/src/main/java/minicraft/entity/mob/Mob.java
@@ -27,6 +27,7 @@ public abstract class Mob extends Entity {
 	int walkTime;
 	public int speed;
 	int tickTime = 0; // Incremented whenever tick() is called, is effectively the age in ticks
+	int noActionTime = 0;
 
 	/**
 	 * Default constructor for a mob.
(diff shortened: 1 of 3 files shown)


src/main/java/minicraft/entity/mob/Mob.java
```diff
@@ -125,6 +127,38 @@ private boolean move(int xd, int yd, boolean changeDir) { // Knockback shouldn't
...
+		return 40;
+	}
+
+	/** @see #checkDespawn() */
+	protected boolean removeWhenFarAway(double distance) {
+		return true;
+	}
+
+	@Override
+	public void checkDespawn() {
```
Makkkkus, line 146: This is a bit misleading because it doesn't check if it is going to despawn. A better name would be handleDespawn.

Makkkkus (approved):

BenCheung0422:
This pull request is ready for reviews.

Makkkkus (approved):

ChristofferHolmesland (approved):

PR #456, opened 2023-03-22
Further resolves gradle run problem

BenCheung0422:
As I test once more, it is found that the problem is not completely resolved. The problem in `FileHandler#listAssets()` is resolved by #421 previously, but the OS-dependent separator problem (zip files must use `/` as the separator but Windows uses `\` as the separator) and directory name conversion problem (directory paths do not directly convert into `String` with a tailing slash, and thus the path name is not recognized as a directory in zip file, which making unexpected files generated in temp zip files) exist. This fixes these problems and makes sure that `gradle run` can work properly.


Litorom (approved):
I’ve had this same issue. Very good

Makkkkus (approved):

PR #457, opened 2023-03-31
Default FPS value

BenCheung0422:
After the change with FPS entry, the default value is always the first element, i.e. 10, which is not expected as a general default value. The default value is now determined by default monitor refresh rate. If the refresh rate is not a multiple of 10, it is subtracted until it is a multiple of 10. If the rate cannot be identified, 60 is used.


ChristofferHolmesland:
Why does it have to be a multiple of 10?

BenCheung0422:
If the values are not separated with a larger value, more time is needed to change to the target FPS and in common cases, multiple of 10 is always generally used.

ChristofferHolmesland (approved):

Makkkkus (approved):

PR #462, opened 2023-04-04
Fix a crash returning item when fishing

BenCheung0422:
This fixes a bug that when a player is fishing and drop the item, open inventory menu or pick up a furniture, an error occurs. This is because the fishing state clean is done only when attacking again or the player is sleeping or swimming, but not for returning the item back to inventory and thus leading a `NullPointerException` for `activeItem` in rendering with `isFishing`.


ChristofferHolmesland (approved):
Approving because it fixes the bug. In the future someone should look at a different way of cleaning up the fishing state so this doesn't happen again in another situation.

Makkkkus:
This does indeed code around the problem. But I think the bugfix is more important as well.

Maybe add this to the changelog?

Makkkkus (approved):

PR #470, opened 2023-04-22
Fire sparks now collide with walls and disappear after a period stopped by walls

BenCheung0422:
Originally, fire sparks pass through walls. but it seems that it would affect the feature of the boss room. The use of the boss room is to block all activities withinn the room interacting outside. Fire sparks also despawn after 50 ticks stopped by walls to prevent duplication and this may lightly increase the difficulty of the boss.


Makkkkus:
Add to changelog, please

Makkkkus (approved):

BenCheung0422:
This change is included in the obsidian knight update which is also included in 2.2.0 changelog, unless there is pre-release changelog.

PR #472, opened 2023-04-29
Reorganise into three subprojects

Makkkkus:
To create a project structure that is scalable enough to encompass a new multiplayer system we split the current grade project into three: `client`, `common`, and `server`. Client and server work as separate programs, while common is a shared layer of code between the two. This should make it possible to build a much more scalable and readable codebase.


BenCheung0422:
I thought we would be doing this in the 2.4 update.

Makkkkus:
It's not a feature so it doesn't matter what update it is in.

BenCheung0422:
Many pull requests still depend on the current project structure, so perhaps after this 2.2 update, then.

Issue #465, opened 2023-04-11
canvas init error on macos x


vvv112:
Unexpected error: Canvas Initialization Failure: The canvas is unable to be initialized.
java.awt.AWTException: Only double buffering is supported
	at java.desktop/sun.java2d.metal.MTLGraphicsConfig.assertOperationSupported(MTLGraphicsConfig.java:274)
	at java.desktop/sun.lwawt.LWComponentPeer.createBuffers(LWComponentPeer.java:515)
	at java.desktop/java.awt.Component$FlipBufferStrategy.createBuffers(Component.java:4116)
	at java.desktop/java.awt.Component$FlipBufferStrategy.<init>(Component.java:4058)
	at java.desktop/java.awt.Component$FlipSubRegionBufferStrategy.<init>(Component.java:4592)
	at java.desktop/java.awt.Component.createBufferStrategy(Component.java:3921)
	at java.desktop/java.awt.Canvas.createBufferStrategy(Canvas.java:201)
	at minicraft.core.Renderer.initScreen(Renderer.java:114)
	at minicraft.core.Game.main(Game.java:80)

vvv112:
java version "20" 2023-03-21
Java(TM) SE Runtime Environment (build 20+36-2344)

zielony12:
they make game for windows 🤦

vvv112:
that is javaVM, is crossplatform

zielony12:
@vvv112  yea but it will crash on other platforms like in this issue

vvv112:
is that fixed now?

Makkkkus:
Yes

PR #474, opened 2023-05-04
Fixed buffering problem on linux and os x

zielony12:
It fixes the canvas initialization issues on Unix-like platforms. Works for me, on linux and should also work for [issue #465](https://github.com/MinicraftPlus/minicraft-plus-revived/issues/465).

### DIFF
--- a/src/main/java/minicraft/core/Renderer.java
+++ b/src/main/java/minicraft/core/Renderer.java
@@ -108,14 +108,7 @@ public class Renderer extends Game {
 		hudSheet = new LinkedSprite(SpriteType.Gui, "hud");
 
 		Initializer.startCanvasRendering();
-		try { // Reference: https://stackoverflow.com/a/61843644.
-			canvas.createBufferStrategy(3, GraphicsEnvironment.getLocalGraphicsEnvironment()
-				.getDefaultScreenDevice()
-				.getDefaultConfiguration()
-				.getBufferCapabilities());
-		} catch (AWTException e) {
-			CrashHandler.crashHandle(e, new ErrorInfo("Canvas Initialization Failure", ErrorInfo.ErrorType.UNEXPECTED, true, "The canvas is unable to be initialized."));
-		}
+		canvas.createBufferStrategy(3);
 
 		canvas.requestFocus();
 	}
--- a/src/main/java/minicraft/core/io/Settings.java
+++ b/src/main/java/minicraft/core/io/Settings.java
@@ -5,6 +5,7 @@ import minicraft.screen.entry.ArrayEntry;
 import minicraft.screen.entry.BooleanEntry;
 
 import java.awt.DisplayMode;
+import java.awt.GraphicsEnvironment;
 import java.util.ArrayList;
 import java.util.HashMap;
 


src/main/java/minicraft/core/Renderer.java
```diff
@@ -236,17 +229,18 @@ private static void renderLevel() {
...
+				for (int x = 0; x <= 280; x++) {
+					screen.render(x, Screen.h - 16 + y * 8, 5, 2, 0, hudSheet.getSheet());
+				}
 			}
 		}
 
 		// Shows active item sprite and name in bottom toolbar.
 		if (player.activeItem != null) {
-			player.activeItem.renderHUD(screen, 10 * 8, Screen.h - 8, Color.WHITE);
+			player.activeItem.renderHUD(screen, 14 * 8, Screen.h - 8, Color.WHITE);
```
Makkkkus, line 241: What's the point of these changes?

zielony12: It brings back the beautiful, full-width and two rows height black rectangle for the status icons. I think that the new bar (that my change overwrites) looks a bit bald and icons without any background looks crappy. It would be so much better to remove the black rectangle at all, however without the rectangle the selected item status looks kind of bad design, and doesn't even fit in minicraft's theme. That is why I decided to change dimensions of the status rectangle. I would really appreciate it, if you let the new code in, or at least make a survey for the community, so they could decide which one fits better.

Makkkkus: The pull request won't be accepted if it contains changes that are outside the scope of the pull request.

zielony12: what about the first, [143d5e7](https://github.com/MinicraftPlus/minicraft-plus-revived/pull/474/commits/143d5e77b1c773ccff9fd8fce2ce2ca9d8e1a7d9) commit?

Makkkkus: First and second commits are okay.

zielony12: Then why the first commit is not applied?

Makkkkus: That is not how pull requests work. It's all or none.

zielony12: I think I have commited a change that overwrites the recent status bar commit

Makkkkus:
Use git reset --hard HEAD~2 and force push that. We don't want messy histories.

Makkkkus (approved):

BenCheung0422 (approved):
The change I have made is to implement VSync, but the trial seemed to have failed. If this change is made, nothing more would be affected. I will reimplement the part of VSync more accurately in the near future.

Litorom:
Umm, I think it work but I noticed via code analysis tool that GraphicsEnvironment is missing so @BenCheung0422 your code may work it’s just missing an import

Makkkkus:
That's not the issue. #465 

In the future. Try to avoid using low-level features in the java language. It almost always breaks something because no-one knows how to do it properly.

PR #489, opened 2023-05-22
Fixes world generation display name input entry rendering problem

### LINKED ISSUE #486: is this supposed to happen?
im just wondering if this is supposed to happen because ive noticed
it a bunch of times while  creating worlds. it probably is intentional but  idk   video:  

 

https://github.com/MinicraftPlus/minicraft-plus-revived/assets/113947915/6368b149-7281-4c24-9c88-00f8acd20bbb

BenCheung0422:
Fixes #486.
The original input entry renders unwisely, which means the entry is rendered with weird displacement, it is unable to check out the texts on the left when the string is longer than the width, also, the position of the entry to displace is not set ideally, it is intended to render precisely but it is not. So, this modifies these behaviours. It is now rendered correctly, and the right arrow on the left also moves according to the displacement of the entry. Also, it is now able to move the entry inspection to the left or to the right by using left and right keys respectively. When changes applied to the entry, the position resets.

### DIFF
--- a/src/main/java/minicraft/screen/WorldGenDisplay.java
+++ b/src/main/java/minicraft/screen/WorldGenDisplay.java
@@ -8,10 +8,13 @@ import minicraft.core.io.Localization;
 import minicraft.core.io.Settings;
 import minicraft.gfx.Color;
 import minicraft.gfx.Font;
+import minicraft.gfx.Point;
 import minicraft.gfx.Screen;
 import minicraft.screen.entry.InputEntry;
 import minicraft.screen.entry.SelectEntry;
 import minicraft.util.Logging;
+import org.jetbrains.annotations.NotNull;
+import org.jetbrains.annotations.Nullable;
 
 import java.nio.file.InvalidPathException;
 import java.nio.file.Paths;
@@ -29,6 +32,7 @@ public class WorldGenDisplay extends Display {
 		if (FileHandler.OS.contains("windows")) {
 			// Reference: https://stackoverflow.com/a/6804755
 			worldNameRegex = "[^<>:\"/\\\\|?*\\x00-\\x1F]+";
+			//noinspection RegExpRepeatedSpace
 			detailedFilenamePattern = Pattern.compile(
 			"# Match a valid Windows filename (unspecified file system).          \n" +
 				"^                                # Anchor to start of string.        \n" +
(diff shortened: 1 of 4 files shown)


src/main/java/minicraft/screen/WorldGenDisplay.java
```diff
@@ -127,19 +131,40 @@ public String getUserInput() {
 				return super.getUserInput().toLowerCase(Localization.getSelectedLocale());
 			}
 
+			private int maxXDisplacement = 0; // Relative maximum rendering displacement.
+			private int renderingX = 0; // Relative rendering displacement.
+
+			{
+				setChangeListener(i -> renderingX = maxXDisplacement = getUserInput().length() > 9 ? -(getUserInput().length()-9) * 8 : 0);
+			}
```
Makkkkus, line 139: Using these brackets means you have designed bad code. This should be done in the constructor or something.

Litorom: Uhh isn’t a constructor a method with the same name as the class? There is clearly a “public worldGenDisplay()” method.

BenCheung0422: This is not how an anonymous class works.

BenCheung0422: > Since they have no name, we can't extend them. For the same reason, anonymous classes cannot have explicitly declared constructors.

Source: https://www.baeldung.com/java-anonymous-classes#1-constructor

Makkkkus: Stop avoiding the question. You know what I mean. Put it in the constructor.

BenCheung0422: That is an anonymous class.

Makkkkus: You mean an inner class? Those can have constructors.

Makkkkus:
This part of the gui system has to be rewritten. It isn't important enough to fix the error with "duct-tape".

BenCheung0422:
This is just a simple fix.

Makkkkus:
It isn't a simple fix. I'm almost certain this will break something.

BenCheung0422:
No, I have tested. This only improves the behaviour of world name input.

BenCheung0422:
The original code does break something, so I am going to implement this in a more general way as this affects more entries and menus in more situations in fact.

PR #491, opened 2023-05-23
Fixes hitbox calculation on entity movement with different speeds

### LINKED ISSUE #246: [BUG] At higher speed of movement, the hitbox of the mobs and that of the player behaves strange
**Describe the bug**
By increasing the speed of an entity or player, the sprite is separated from the hitbox

**To Reproduce**
1. Drink a speed potion (or just increase the speed of the player)
2. Try to hit an enemy or a tile (you will notice that the player's sprite is further away from the hitbox, also and observe that the distance between the player's sprite and the tile increases, when trying to attack an enemy, most of the time it could fail the hit)

**Expected behavior**
The hitbox of the mobs (including the player) should be centered with the sprite

**Screenshots**
### Enemy hit:
test

### Tile distance, move speed = 2:
dung

### Tile distance, move speed = 3 or more:
dung

**Desktop (please complete the following information):**
 - OS: [Windows 10, It is more likely to happen in other systems]
 - Version [all 2.0.1 - 2.0.7]

**Additional context**
Apparently the hitbox moves a lot to the right

BenCheung0422:
Fixes #246.
When the speed of entity is not 1, the hitbox calculation encounters an unexpected problem, and the result does not meet the actual movement. This also fixes a potential problem that the entity hitbox could run into weird behavior at higher movement, i.e., hitbox teleporting. This way calculates the hitbox 1 by 1 instead of the whole.
Actually, there is another approach that might involve less looping by calculating the minimum distance from the entity hit box to the closest collidable hit box. However, that might be more complex in implementation.


src/main/java/minicraft/entity/Axis2D.java
```diff
@@ -0,0 +1,12 @@
...
+public enum Axis2D {
+	X, Y;
+
+	public Direction getDirection(int displacement) {
+		if (this == X)
+			return Direction.getDirection(displacement, 0);
+		else
+			return Direction.getDirection(0, displacement);
+	}
+}
```
Makkkkus, line 12: This is a case where it would be better to use a function instead of a class. Let's try to avoid the excessive use of OOP.

src/main/java/minicraft/entity/Entity.java
```diff
@@ -132,74 +134,127 @@ public boolean move(int xd, int yd) {
...
 
-		// Finally, the entity moves!
-		x += xd;
-		y += yd;
-
-		return true; // the move was successful.
+		return successful;
 	}
 
 	/** Checks if the entity is able to naturally be despawned in general conditions. Handles (despawns) if true. */
```
Makkkkus, line 260: There must be a better way of doing this. Manually calculating the bounds and etc of the entity seemes a little bit strange to put in one single method, let alone the same class.

Makkkkus (changes requested):

src/main/java/minicraft/entity/Entity.java
BenCheung0422: Right, but the manual calculation is still needed to make sure these are working correctly. Hmm, probably rectangle interception could be involved to simplify boundary calculation.

BenCheung0422: If the movement is extremely small here, the calculation is quite linear as loops are always skipped. Besides, if the movement is not small, looping calculation is always needed to search for hit box interactions.

Makkkkus (approved):
Pretty good

ChristofferHolmesland (approved):

PR #495, opened 2023-05-25
Update splash screen coloring

BenCheung0422:
First, there is only black background used in the current splash screen. Second, the speed of the fading is now constant and not depending on the looping speed. Also, the fading calculation problem on the alpha channel is now fixed, so it would now fade as expected. The faded in screen now holds at least 180ms before fading out even if background execution is finished. The whole fading process: fading in (306ms) -> holding (at 180ms least until background process finishes) -> fading out (360ms).

### DIFF
--- a/src/main/java/minicraft/core/Initializer.java
+++ b/src/main/java/minicraft/core/Initializer.java
@@ -188,7 +188,7 @@ public class Initializer extends Game {
 		public Thread renderer = new Thread(() -> {
 			long lastTick = System.nanoTime();
 			do {
-				if (System.nanoTime() - lastTick >= 1E7) { // 10ms/tick
+				if (System.nanoTime() - lastTick >= 6E6) { // 6ms/tick
 					repaint();
 					lastTick = System.nanoTime();
 				}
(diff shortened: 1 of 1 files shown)


Litorom:
Can you add a gif of the new splash, I can’t download/run code on an iPad. (I wish I could tho)

BenCheung0422:
https://github.com/MinicraftPlus/minicraft-plus-revived/assets/74168521/b3412399-d279-4f2c-b9a6-1eae577ed7af

BenCheung0422:
There is a player mentioned that:

> I don't like the rainbow color effect of the startup, idk why, It's maybe because of the color saturation
> I don't like the new rainbow startup
> i feel like it doesn't fit at all with the game
> a black startup with the logo slowly appearing and then shinning would be better in my opinion
> (and after shinning slowly disappearing again to lead to the main menu)

BenCheung0422:
Also, this includes the fixes on calculations that the algorithms do not meet the actual results.

BenCheung0422:
Why is it pointless and should not be included?

BenCheung0422:
I have entirely simplified the splash screen. See the updated description.

https://github.com/MinicraftPlus/minicraft-plus-revived/assets/74168521/f38d11b7-23a1-4094-ad01-b98540168d00

Makkkkus:
I don't trust that this works cross-platform.

1. bloat
2. keep to the roadmap

BenCheung0422:
What do you mean by "works cross-platform"?

PR #497, opened 2023-06-07
Fix items breaking [PlantTile]s

Dioarya:
Pull request to fix items from breaking [PlantTile]s.

### DIFF
--- a/src/main/java/minicraft/level/tile/farming/PlantTile.java
+++ b/src/main/java/minicraft/level/tile/farming/PlantTile.java
@@ -21,13 +21,8 @@ public class PlantTile extends FarmTile {
     @Override
     public void steppedOn(Level level, int xt, int yt, Entity entity) {
 		if (entity instanceof ItemEntity) return;
-		// Very low chance because steppedOn gets called every tick of movement on the tile.
 		if (random.nextInt(100) != 0) return;
 		harvest(level, xt, yt, entity);
-		if (random.nextInt(10) != 0)
-			level.setTile(xt, yt, Tiles.get("Farmland"));
-		else
-			level.setTile(xt, yt, Tiles.get("Dirt"));
     }
 
     @Override
@@ -81,6 +76,9 @@ public class PlantTile extends FarmTile {
 		// Play sound.
 		Sound.play("monsterhurt");
 
-        level.setTile(x, y, Tiles.get("Dirt"));
+        if (random.nextInt(7) == 0)
+			level.setTile(x, y, Tiles.get("Farmland"));
+		else
+			level.setTile(x, y, Tiles.get("Dirt"));
     }
 }
--- a/src/main/java/minicraft/level/tile/farming/PlantTile.java
+++ b/src/main/java/minicraft/level/tile/farming/PlantTile.java
@@ -3,6 +3,7 @@ package minicraft.level.tile.farming;
 import minicraft.core.io.Sound;
 import minicraft.entity.Direction;
 import minicraft.entity.Entity;
+import minicraft.entity.ItemEntity;
 import minicraft.entity.mob.Mob;
 import minicraft.entity.mob.Player;
 import minicraft.item.Items;
@@ -20,6 +21,7 @@ public class PlantTile extends FarmTile {
     @Override
     public void steppedOn(Level level, int xt, int yt, Entity entity) {
         super.steppedOn(level, xt, yt, entity);
+		if (entity instanceof ItemEntity) return;
         harvest(level, xt, yt, entity);
     }
 
--- a/src/main/java/minicraft/level/tile/farming/PlantTile.java
+++ b/src/main/java/minicraft/level/tile/farming/PlantTile.java
@@ -20,9 +20,13 @@ public class PlantTile extends FarmTile {
 
     @Override
     public void steppedOn(Level level, int xt, int yt, Entity entity) {
-        super.steppedOn(level, xt, yt, entity);
 		if (entity instanceof ItemEntity) return;
+		// Very low chance because steppedOn gets called every tick of movement on the tile.
+		if (random.nextInt(200) != 0) return;
         harvest(level, xt, yt, entity);
+		if (random.nextInt(20) == 0)
+			level.setTile(xt, yt, Tiles.get("Dirt"));
+		super.steppedOn(level, xt, yt, entity);
     }
 
     @Override
@@ -76,6 +80,6 @@ public class PlantTile extends FarmTile {
 		// Play sound.
 		Sound.play("monsterhurt");
 
-        level.setTile(x, y, Tiles.get("Dirt"));
+        level.setTile(x, y, Tiles.get("Farmland"));
     }
 }
--- a/src/main/java/minicraft/level/tile/farming/PlantTile.java
+++ b/src/main/java/minicraft/level/tile/farming/PlantTile.java
@@ -22,11 +22,12 @@ public class PlantTile extends FarmTile {
     public void steppedOn(Level level, int xt, int yt, Entity entity) {
 		if (entity instanceof ItemEntity) return;
 		// Very low chance because steppedOn gets called every tick of movement on the tile.
-		if (random.nextInt(200) != 0) return;
-        harvest(level, xt, yt, entity);
-		if (random.nextInt(20) == 0)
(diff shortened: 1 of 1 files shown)


BenCheung0422:
What if we pull the super method call down and copy some if statements from Farmland to this?

Dioarya:
I'm sorry, I don't really get what you mean by that. I'm very new to java programming

Dioarya:
Maybe we don't need `FarmTile.super` in `PlantTile.steppedOn` because it wouldn't make sense for a chance to turn to dirt from `FarmTile.super` if we're going to turn the tile to a dirt tile with `PlantTile.harvest` right after it, regardless if it is an item or not. Should I add this to a commit?

BenCheung0422:
Good point. However, the original expected mechanics is when there is an entity stepping on a plant tile, the tile will be immediately harvested and having a chance changing back in a dirt tile. I think keeping the idea is fine, but I think changing 100% being harvested into having a chance being harvested would be better.

Dioarya:
I think I get you, so when `PlantTIle` is being stepped on there's a chance for the `PlantTile` to turn back to `FarmTile` (being harvested) and there's a second chance on top of that to turn into a dirt tile?

BenCheung0422:
Yes. By the way, it would be better to put description as being a pull request.

Dioarya:
Btw do you know what `level.getData` means?

BenCheung0422:
It gets the tile data by the corresponding level and coordinates.

src/main/java/minicraft/level/tile/farming/PlantTile.java
```diff
@@ -74,6 +80,6 @@ protected void harvest(Level level, int x, int y, Entity entity) {
 		// Play sound.
 		Sound.play("monsterhurt");
 
-        level.setTile(x, y, Tiles.get("Dirt"));
+        level.setTile(x, y, Tiles.get("Farmland"));
```
BenCheung0422, line 83: I would like to keep the original mechanics.

src/main/java/minicraft/level/tile/farming/PlantTile.java
```diff
@@ -19,8 +20,13 @@ protected PlantTile(String name) {
...
     @Override
     public void steppedOn(Level level, int xt, int yt, Entity entity) {
-        super.steppedOn(level, xt, yt, entity);
+		if (entity instanceof ItemEntity) return;
+		// Very low chance because steppedOn gets called every tick of movement on the tile.
+		if (random.nextInt(200) != 0) return;
         harvest(level, xt, yt, entity);
+		if (random.nextInt(20) == 0)
+			level.setTile(xt, yt, Tiles.get("Dirt"));
+		super.steppedOn(level, xt, yt, entity);
```
BenCheung0422, line 29: This is quite redundant as changing into a dirt tile is handled in the super invoke.

src/main/java/minicraft/level/tile/farming/PlantTile.java
```diff
@@ -19,8 +20,13 @@ protected PlantTile(String name) {
 
     @Override
     public void steppedOn(Level level, int xt, int yt, Entity entity) {
-        super.steppedOn(level, xt, yt, entity);
+		if (entity instanceof ItemEntity) return;
+		// Very low chance because steppedOn gets called every tick of movement on the tile.
+		if (random.nextInt(200) != 0) return;
```
BenCheung0422, line 25: I think the chance is too low compared to the chance changing back into a dirt tile.

BenCheung0422 (changes requested):

src/main/java/minicraft/level/tile/farming/PlantTile.java
Dioarya: `FarmTile.steppedOn`'s chances is so low to compensate being called every tick you are moving on a `FarmTile`. But this is called 1 time when the if statement above this doesn't return out of the function, which makes the tile gets harvested.

BenCheung0422: This is called only when the player is moving.

src/main/java/minicraft/level/tile/farming/PlantTile.java
Dioarya: I guess this does affect when `PlantTile.hurt` is called. Alright.

src/main/java/minicraft/level/tile/farming/PlantTile.java
Dioarya: Yeah that's why I said every tick that you are moving on the tile...

BenCheung0422: Then, this makes a duplication. At least removing either one of these statements.

src/main/java/minicraft/level/tile/farming/PlantTile.java
Dioarya: Okay, what do you think it should be? Keep in mind that this is called every tick that you are moving on the tile, with 60 ticks per second.

BenCheung0422: It is not every tick. If a player is passing through a farmland straight, this is called only 16 times in total with normal speed.

BenCheung0422:
I would like to tweak this mechanics: the chance of changing back into a dirt (high) or a farmland (low) is handled in `#harvest`; there is a chance to harvest each time stepping on the tile; the inherit call of `FarmTile#steppedOn` should be removed as it is duplicated among these methods; item entity check remains kept.

src/main/java/minicraft/level/tile/farming/PlantTile.java
Dioarya: I just tested it with `Logging.TILES.debug("stepping" + System.currentTimeMillis());` and after a second theres 60 lines with every datapoint having ~16ms delay from each other though... I had a feeling of this because on `FarmTile.steppedOn` it's 1/60 chance and it took around 1 second to turn the `FarmTile` to dirt.

Dioarya: > I think the chance is too low compared to the chance changing back into a dirt tile.

I will change it to 100 and 10 respectively.

Dioarya:
> I would like to tweak this mechanics: the chance of changing back into a dirt (high) or a farmland (low) is handled in `#harvest`; there is a chance to harvest each time stepping on the tile; the inherit call of `FarmTile#steppedOn` should be removed as it is duplicated among these methods; item entity check remains kept. It seems that the original author has not successfully made this mechanics as what he wanted.

okay

Dioarya:
> I would like to tweak this mechanics: the chance of changing back into a dirt (high) or a farmland (low) is handled in `#harvest`; there is a chance to harvest each time stepping on the tile; the inherit call of `FarmTile#steppedOn` should be removed as it is duplicated among these methods; item entity check remains kept. It seems that the original author has not successfully made this mechanics as what he wanted.

Wait... Changing `PlantTile.harvest`'s code affects how `PlantTile.hit` works too... how shall we fix this?

BenCheung0422:
Yes, it does. However, I think this mechanics is good for both hitting and stepping on.

Dioarya:
Okay, is everything ready to merge?

BenCheung0422 (approved):

Makkkkus:
This was already done by the PlantTile class.

Dioarya:
It hasn't...

BenCheung0422:
@Makkkkus 

https://github.com/MinicraftPlus/minicraft-plus-revived/assets/74168521/3482f459-0e50-4e1c-b4f8-a2946d822bc4

Makkkkus:
This is done by farm tile as seen here: https://github.com/MinicraftPlus/minicraft-plus-revived/blob/22b52dbf39c70c0ef15ee486a03d5c0aa3cb6fa4/src/main/java/minicraft/level/tile/farming/FarmTile.java#L56

BenCheung0422:
This only returns and breaks *the method it is currently executing*, so the line your mentioned is not related to the call from a subclass. In a subclass, it only immediately finishes the call to this inherited method. This is not a valid argument,

Makkkkus:
It was working fine before, so I don't know who broke this.

PR #454, opened 2023-03-19
Reorganized debugging options

BenCheung0422:
*This pull request depends on #456*
Running the game with `java.exe` would have already been intended as *debugging*, so a simple (redundant/meaningless) `debug` flag should not be used in logging. Instead, separated `log-level` flag is used for logging level details, and all the debugging key bindings are changed to stick with `F3`, the major debugging key. Additionally, as #429 has added debug locale, this also adds a flag "debug-locale" for it.
- 2 key mappings are removed: `SURVIVAL=debug` and `CREATIVE=debug`.
- Complex compound (multiple keys) key binding support replacing dual binding support of `SHIFT`, `CTRL` or `ALT`.
- `Game#debug` field is removed and `Logging#logLevel` (for logging level details) and `Localization#isDebugLocaleEnabled` are added.
- Redundant `if` expressions in `if (Game.debug) Logging.debug();` expressions are simplified.
- 2 new key bindings are added:
  - `F3+Y` - Moving up a level.
  - `F3+H` - Moving down a level.
- Key binding changes:
  - `CTRL+P -> F3+L` - Printing players on all levels.
  - `{1|2|3|4} -> F3+T+{1|2|3|4}` - Setting the current daytime. 1: morning, 2: day, 3: evening, 4: night.
  - `SURVIVAL=debug -> F3+F4+1` - Setting game mode to survival.
  - `CREATIVE=debug -> F3+F4+2` - Setting game mode to creative.
  - `SHIFT+T -> F3+F4+2` - Setting game mode to score.
  - `CTRL+T -> F3+SHIFT+T` - Setting the time in score mode to 5 seconds.
  - `SHIFT+0 -> F3+S+0` - Resetting the game tick speed.
  - `SHIFT+= -> F3+S+=` - Increasing the game tick speed.
  - `SHIFT+- -> F3+S+-` - Decreasing the game tick speed.
  - `CTRL+H -> F3+H` - Decreasing health value by 1.
  - `CTRL+B -> F3+B` - Decreasing hunger value by 1.
  - `0 -> F3+M+0` -> Resetting player moving speed.
  - `= -> F3+M+=` -> Increasing player moving speed.
  - `- -> F3+M+-` -> Decreasing player moving speed.
  - `SHIFT+U -> F3+U` -> Setting the current standing tile as stairs up.
  - `SHIFT+D -> F3+D` -> Setting the current standing tile as stairs down.
  - `SHIFT+P -> F3+P` -> Removing all potion effects acting on player.
[...]

PR #501, opened 2023-06-16
Removes redundant debugging level change hotkeys

BenCheung0422:
As this should have been replaced by #454, but this seems that it has been neglected upon the change.
Corresponding key binding changes:
- `SHIFT+W` -> `F3+Y`
- `SHIFT+S` -> `F3+H`


Makkkkus:
I think we should go back to using the old debugging buttons instead. Might revert #454.

But to be honest. It doesn't really matter...

PR #502, opened 2023-06-16
Makes debugging panel unable to flick

BenCheung0422:
As there are key bindings binding to F3, when those key bindings are being inputted, the debugging panel is also triggered, which is not intended. When pressing F3, it should be only triggered once, which is when the F3 key is released. This resolves these problems.


Makkkkus:
Bloat. Feature creep. Whatever you may call it. You're also probably using a solution that is much more complicated than just using a different method for checking input. I will not add this.

Makkkkus:
We will eventually use a different input system in the future, so bring this up then instead.

PR #503, opened 2023-06-17
Fix problems caused by a revert in FPS entry change

BenCheung0422:
A change in FPS entry made by #425 has been reverted by https://github.com/MinicraftPlus/minicraft-plus-revived/commit/9de1e2a200cec588d80e21d9a2bd4b3df976104d, but the change has already been proposed in 2.2.0-dev3, which may cause potential compatibility problem, so this fixes this problem, also redoes a change by #457 and a change from interval of 1 to 10 in FPS setting values.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -6,7 +6,7 @@ plugins {
 }
 
 
-project.version = '2.2.0-dev3'
+project.version = '2.2.0-dev4'
 group = "java.minicraft"
 archivesBaseName = "minicraft-plus"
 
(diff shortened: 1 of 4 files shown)


Makkkkus:
You're fixing problems that you caused yourself with the major change you made yourself. Just change your settings file.

BenCheung0422:
Em...that has been released in a pre-release. Aren't we ensuring the compatibility in each pre-release?

Makkkkus:
Nope. They're pre-releases. Or we can just delete it.

BenCheung0422:
Delete what?

PR #504, opened 2023-06-19
Adds a Debug Panel

BenCheung0422:
This adds a debug panel and removes all in-world debugging key bindings. This could make the key bindings cleaner and neat. Instead, a list entry menu (triggered by F4) is used and searching is available. (No localization) This would be a way to organize all the debugging functions and features without a lot of (confusing or messy) key bindings.

### DIFF
--- a/src/main/java/minicraft/core/Renderer.java
+++ b/src/main/java/minicraft/core/Renderer.java
@@ -28,7 +28,6 @@ import minicraft.screen.LoadingDisplay;
 import minicraft.screen.Menu;
 import minicraft.screen.QuestsDisplay;
 import minicraft.screen.RelPos;
-import minicraft.screen.ResourcePackDisplay;
 import minicraft.screen.TutorialDisplayHandler;
 import minicraft.screen.entry.ListEntry;
 import minicraft.screen.entry.StringEntry;
@@ -37,10 +36,8 @@ import minicraft.util.Quest.QuestSeries;
 
 import javax.imageio.ImageIO;
 
-import java.awt.AWTException;
 import java.awt.Canvas;
 import java.awt.Graphics;
-import java.awt.GraphicsEnvironment;
 import java.awt.geom.AffineTransform;
 import java.awt.image.AffineTransformOp;
 import java.awt.image.BufferStrategy;
--- a/src/main/java/minicraft/core/Updater.java
+++ b/src/main/java/minicraft/core/Updater.java
@@ -8,6 +8,7 @@ import minicraft.level.Level;
 import minicraft.level.tile.Tile;
 import minicraft.level.tile.Tiles;
 import minicraft.saveload.Save;
+import minicraft.screen.DebugPanelDisplay;
 import minicraft.screen.Display;
 import minicraft.screen.EndGameDisplay;
 import minicraft.screen.LevelTransitionDisplay;
@@ -90,15 +91,6 @@ public class Updater extends Game {
 	// VERY IMPORTANT METHOD!! Makes everything keep happening.
 	// In the end, calls menu.tick() if there's a menu, or level.tick() if no menu.
 	public static void tick() {
-
-		// Quick Level change: move the player for -1, or 1 levels
-		if (isMode("minicraft.settings.mode.creative") && input.getKey("SHIFT-S").clicked ) {
-			Game.setDisplay(new LevelTransitionDisplay(-1));
-
-		} else if (isMode("minicraft.settings.mode.creative") && input.getKey("SHIFT-W").clicked ){
-			Game.setDisplay(new LevelTransitionDisplay(1));
-		}
-
 		if (input.getKey("FULLSCREEN").clicked) {
 			Updater.FULLSCREEN = !Updater.FULLSCREEN;
 			Updater.updateFullscreen();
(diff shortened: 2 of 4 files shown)


Makkkkus:
No new debug panel is needed. The argumens I presented in #502 still apply. This is bloat, and I would prefer going back to the way it was before #454 (although not all changes are bad).

BenCheung0422:
However, I think there might be too many key bindings and that might make confusion.

Makkkkus:
It doesn't matter. They're only possible to use when using debug mode.

BenCheung0422:
It can still be a problem as sometimes lookup for key bindings are still required if the key bindings are forgotten.

Makkkkus:
That's why we have our [wiki](https://github.com/MinicraftPlus/minicraft-plus-revived/wiki/Debug-powers). Using F3 or F4 won't change that.

BenCheung0422:
And a in-game instant lookup. F3 only triggers the debugging information panel, but this F4 triggers an action panel.

Makkkkus:
Yeah, it's not needed.

BenCheung0422:
There are 23 associated debugging key bindings, if every time using the bindings requires lookup, using built-in interface with only 1 key would be better than opening up the wiki page on browser and find line by line the key bindings required.

Makkkkus:
I'll reopen this on the condition that you readd the --debug launch argument.

BenCheung0422:
I have re-added `Game#debug` and `--debug` flag. The corresponding codes that was relying on that field have also recovered. I am not sure if the detailed debugging flags `--debug-*` should be enabled only when `--debug` is also enabled.

src/client/java/minicraft/level/tile/StairsTile.java
```diff
@@ -40,7 +40,7 @@ public boolean mayPass(Level level, int x, int y, Entity e) {
 	public boolean interact(Level level, int xt, int yt, Player player, Item item, Direction attackDir) {
 		super.interact(level, xt, yt, player, item, attackDir);
 
-		// Makes it so you can remove the stairs if you are in creative and debug mode.
+		// Makes it so you can remove the stairs if you are in creative mode.
 		if (item instanceof PowerGloveItem && Game.isMode("minicraft.settings.mode.creative")) {
```
Makkkkus, line 44: Are you able to place stairs in creative?

src/client/java/minicraft/network/Analytics.java
```diff
@@ -52,6 +53,7 @@ public enum Analytics {
 
 	@Nullable public Future<HttpResponse<Empty>> ping() { return ping(1); }
 	@Nullable public Future<HttpResponse<Empty>> ping(int value) {
+		if (Game.debug) return null;
 		final String url = "https://pingdat.io?t="+token+"&v="+value;
 
 		return Unirest.get(url).asEmptyAsync(new Callback<Empty>() {
```
Makkkkus, line 59: I'm unsure how legal this class in in the EU. It's probably legal in the USA, since it has been implemented, but I am pretty sure this is a breach of the GDPR act.

src/client/java/minicraft/level/tile/StairsTile.java
BenCheung0422: We could add the ability then.

src/client/java/minicraft/network/Analytics.java
BenCheung0422: I think we should have privacy policy and Terms of Services or EULA for this.

src/client/java/minicraft/level/tile/StairsTile.java
Makkkkus: Did you or did you not remove the debug option?

BenCheung0422: There was a debug check for this, but I think it should be able also without debug.

Makkkkus: You shouldn't be able to destroy stairs in non-debug mode as you can only place them with the debug menu. Don't add a new item!

src/client/java/minicraft/network/Analytics.java
Makkkkus: Doesn't matter. Still illegal.

src/client/java/minicraft/level/tile/StairsTile.java
BenCheung0422: Ok.

src/client/java/minicraft/network/Analytics.java
BenCheung0422: I am not sure about that, but that was existed for a long time.

BenCheung0422: How should we handle this then?

BenCheung0422:
There could be a function to area set tiles with using menus (inputs).

Makkkkus:
I don't understand what you mean?

BenCheung0422:
When making maps, special actions to changes tiles and entities are required. Testing some features may also require large range tiles.

Makkkkus:
Still don't understand. Creating maps?

BenCheung0422:
Customizing worlds in creative mode. Although there are currently no trends in making maps and publishing maps because there is lack of this kind of functionality, testing some features is still requiring an environmental setup and that might require major change to the testing world. Having this kind of debugging or creative mode exclusive functions could reduce the work required and make convenience. Some of this kind of functions I can think of are tile setting, region filling, entity summoning, potion effect modifying, etc.

Makkkkus:
What do you want to do?

BenCheung0422:
Add a debug function entry in the panel and use menus to do the actions.

Makkkkus:
We already have creative mode. Don't destroy that.

BenCheung0422:
It also adds another function to creative mode, like Minecraft's `fill` command.

Makkkkus:
Okay. That's something I see the use for.

BenCheung0422:
There are several functions I want to add for the panel (might be enabled in creative mode?):
- regional tile filling - implementation with using a sub-menu of tile list or input entry for tile selection and another 2 sets of input entries specifying the region.
- tile setting - specifying a tile with positions and modifying the tile data.
- entity killing - killing entities by conditions.
- item giving - giving an item with using an input entry and may allow modifying item data.
- status effect modifying - adding or removing player's status effects (potion effects).
- inventory clearing - clearing the whole inventory.
- day locking or time locking - locking the daytime and day-night cycle.
- entity summoning - summoning a or a number of specific entity(s) without a spawner (softly or forcedly).

BenCheung0422:
Right, since they are debug functions. Then what about the functions mentioned to be added?

Makkkkus:
All of them are fine. It's the implementation that matters in this case.

BenCheung0422:
However, I am not sure how should I do for tile selection. Tile list menu could be intuitive, but an input entry could be convenient without scrolling.

BenCheung0422:
We might change some of them into string commands if we need to. Also, the implementations of the commands I proposed previously have been done.

ChristofferHolmesland:
I have reviewed the PR, but I don't approve of the changes as explained in Discord.

PR #324, opened 2022-04-25
Make book writable

BenCheung0422:
Fixes #319

PR #509, opened 2023-06-22
Add item description

BenCheung0422:
This reserves the feature of item description, which might be useful when there are colorable items and more items based on non-name-indicated attributes. Taken from #324.


src/main/java/minicraft/screen/PlayerInvDisplay.java
```diff
@@ -7,54 +7,77 @@
...
+import minicraft.screen.entry.StringEntry;
 
 public class PlayerInvDisplay extends Display {
 
 	private static final int padding = 10;
 
 	private final Player player;
 
+	private String itemDescription = "";
+	private Menu.Builder descriptionMenuBuilder;
```
Makkkkus, line 25: You don't need this.

Makkkkus (changes requested):

src/main/java/minicraft/screen/PlayerInvDisplay.java
BenCheung0422: Why we don't need it? It is a dynamic menu size.

Makkkkus: Just don't keep the builder stored.

BenCheung0422: The builder is always stored every time the frame sizes are modified. It also appears in other places as commonly.

Makkkkus (approved):

ChristofferHolmesland (approved):

PR #510, opened 2023-06-22
Player inventory item entry ordering and stacking functions

BenCheung0422:
As there has already been entry limit on the player inventory, item stack operations should be introduced to make support to the inventory limitation. This adds functionalities of item entry ordering (by shifting entries/stacks) and stacking (stackable items that have spare space from maximum stacking size). The corresponding key bindings are `SHIFT+CURSOR-[UP|DOWN]` and `CTRL+CURSOR-[UP|DOWN]`. This also fixes the problems that stackable items always adding at the end of the inventory even it has indicated to other slots, found by the item re-adding feature. Currently, if the target slot is not specified, all items will be attempted to be added at the end of the inventory; otherwise, if the item is stackable, the items (occupying 1 slot only if `singleton`) are attempted to be added to the target slot, or else (if `singleton` is false (in player inventory menu or container menu), then first attempt to add to the following next slot) the remaining items are attempted to be added at the end of the inventory. Redundant numeral checks are also reduced from `Inventory#add` and a nullable item object is returned for the check. When holding `CTRL`, the added items attempt to add as the unspecified slot.


BenCheung0422:
Related post: https://discord.com/channels/280723930942013440/1142909671846445076

BenCheung0422:
This feature has also been mentioned in `possible future projects.txt`, "Inventory Sorting", but without the tabs/categories.

Litorom:
Denied

PR #514, opened 2023-06-23
Make FPS setting having an interval of 10 in values

BenCheung0422:
As described, this redoes changes made by #503 except the change to file loading. The change has original been made by #457, which is from interval of 1 to 10 in FPS setting values. It could reduce the time in changing the setting value as multiples of 10 are commonly used values. Also, the default FPS setting values have been doubled, because a doubled FPS is more optimized in rendering preventing most lagging by frame loss.

### DIFF
--- a/src/client/java/minicraft/core/io/Settings.java
+++ b/src/client/java/minicraft/core/io/Settings.java
@@ -14,7 +14,7 @@ public final class Settings {
 	private static final HashMap<String, ArrayEntry<?>> options = new HashMap<>();
 
 	static {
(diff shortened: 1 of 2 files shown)


Makkkkus:
I've already said no to this.

BenCheung0422:
Why is that bad?

Makkkkus:
Some people (like myself) have monitors that have refresh rates like 74 hz. Your proposal would fuck absoutely anyone that has a refresh rate that isn't a multiple of 10. In addition 60 fps is the standard. Remeber that we are targeting low-end devices.

BenCheung0422:
Alright.

PR #517, opened 2023-06-24
Fix Gradle run problem

BenCheung0422:
As there is a change to the project structure, if it is running without further configuration, server would also be executed, which is not intended. This resolves this problem by overriding the main class of client.


Makkkkus:
You set the class it should run in the IDE.

BenCheung0422:
Then `gradle run` is no longer a valid command?

Makkkkus:
Use command arguments.

BenCheung0422:
README.md should be updated for that.

PR #518, opened 2023-06-24
Create SpriteLinkBuilder

BenCheung0422:
As the current `LinkedSprite` is fully configurable in runtime, which is not really suitable to be a carrier. A class `LinkedSprite$SpriteLinkBuilder` is added as a builder for `LinkedSprite`.

### DIFF
--- a/src/client/java/minicraft/core/Renderer.java
+++ b/src/client/java/minicraft/core/Renderer.java
@@ -52,6 +52,7 @@ import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
 import java.util.Objects;
+import java.util.function.Supplier;
 
 public class Renderer extends Game {
 	private Renderer() {
@@ -76,7 +77,7 @@ public class Renderer extends Game {
 
 	private static int potionRenderOffset = 0;
 
-	private static LinkedSprite hudSheet;
+	private static Supplier<MinicraftImage> hudSheet;
 
 	public static MinicraftImage loadDefaultSkinSheet() {
 		MinicraftImage skinsSheet;
@@ -102,7 +103,7 @@ public class Renderer extends Game {
 
 		image = new BufferedImage(WIDTH, HEIGHT, BufferedImage.TYPE_INT_RGB);
 		screen.pixels = ((DataBufferInt) image.getRaster().getDataBuffer()).getData();
-		hudSheet = new LinkedSprite(SpriteType.Gui, "hud");
+		hudSheet = () -> Renderer.spriteLinker.getSheet(SpriteType.Gui, "hud");
 
 		Initializer.startCanvasRendering();
 		canvas.createBufferStrategy(3);
@@ -227,7 +228,7 @@ public class Renderer extends Game {
 		// This draws the black square where the selected item would be if you were holding it
 		if (!isMode("minicraft.settings.mode.creative") || player.activeItem != null) {
 			for (int x = 10; x < 26; x++) {
-				screen.render(x * 8, Screen.h - 8, 5, 2, 0, hudSheet.getSheet());
+				screen.render(x * 8, Screen.h - 8, 5, 2, 0, hudSheet.get());
 			}
 		}
 
@@ -247,7 +248,7 @@ public class Renderer extends Game {
 				else
 					Font.drawBackground("	x" + ac, screen, 84, Screen.h - 16);
 				// Displays the arrow icon
-				screen.render(10 * 8 + 4, Screen.h - 16, 4, 1, 0, hudSheet.getSheet());
+				screen.render(10 * 8 + 4, Screen.h - 16, 4, 1, 0, hudSheet.get());
 			}
 		}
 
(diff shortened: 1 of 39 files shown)


src/client/java/minicraft/core/Game.java
```diff
@@ -87,6 +88,7 @@ public static void main(String[] args) {
 		Analytics.GameStartup.ping();
 
 		input = new InputHandler(Renderer.canvas);
+		Renderer.spriteLinker = new SpriteLinker();
```
Makkkkus, line 91: Perhaps this should be in `Renderer.initScreen` instead?

src/client/java/minicraft/core/Renderer.java
```diff
@@ -102,7 +103,7 @@ public static void initScreen() {
 
 		image = new BufferedImage(WIDTH, HEIGHT, BufferedImage.TYPE_INT_RGB);
 		screen.pixels = ((DataBufferInt) image.getRaster().getDataBuffer()).getData();
-		hudSheet = new LinkedSprite(SpriteType.Gui, "hud");
+		hudSheet = () -> Renderer.spriteLinker.getSheet(SpriteType.Gui, "hud");
```
Makkkkus, line 106: Please avoid complicated expressions like this. It would be better to just store the sheet in memory.

src/client/java/minicraft/core/Renderer.java
```diff
@@ -544,24 +545,24 @@ private static void renderFocusNagger() {
...
-			screen.render(xx + x * 8, yy - 8, 1, 6, 0, hudSheet.getSheet()); // ...Top part
-			screen.render(xx + x * 8, yy + 8, 1, 6, 2, hudSheet.getSheet()); // ...Bottom part
+			screen.render(xx + x * 8, yy - 8, 1, 6, 0, hudSheet.get()); // ...Top part
+			screen.render(xx + x * 8, yy + 8, 1, 6, 2, hudSheet.get()); // ...Bottom part
 		}
 		for (int y = 0; y < h; y++) {
-			screen.render(xx - 8, yy + y * 8, 2, 6, 0, hudSheet.getSheet()); // ...Left part
-			screen.render(xx + w * 8, yy + y * 8, 2, 6, 1, hudSheet.getSheet()); // ...Right part
+			screen.render(xx - 8, yy + y * 8, 2, 6, 0, hudSheet.get()); // ...Left part
+			screen.render(xx + w * 8, yy + y * 8, 2, 6, 1, hudSheet.get()); // ...Right part
```
Makkkkus, line 560: Is is a clever an logical name change.

Makkkkus (changes requested):
Try to reduce the amount of SpriteLinker classes. It would be better to group many items togheter in a sprite sheet.

src/client/java/minicraft/core/Game.java
BenCheung0422: It must be created before the creations of tiles and items.

Makkkkus: Put it before tiles and items are created then.

src/client/java/minicraft/core/Renderer.java
BenCheung0422: The sheet object is changed in the linker when changes applied in resources.

src/client/java/minicraft/core/Game.java
BenCheung0422: Resource packs also require linker to register sheets.

Makkkkus: Put it in the constructor of Renderer then.

src/client/java/minicraft/core/Renderer.java
Makkkkus: Could you profile the performance impact of before and after this pull request?

src/client/java/minicraft/core/Game.java
BenCheung0422: Ok.

src/client/java/minicraft/core/Renderer.java
BenCheung0422: This change (on this line) relates to the mechanics of `LinkedSprite`. A `LinkedSprite` updates on demand (when changes of resources applied, valid `LinkedSprite`s update), but it is expected that a `Sprite` is used instead of a sheet.

BenCheung0422: In thread time, it changes from around 920±50 ms/s to around 300±150 ms/s on main thread with the same (initialized) world.
In memory (allocated bytes) per thread, it changes from 9±0.4 MB to around 9.8±0.4 MB on main thread.
These are measured with simple movements and at rest on the same (initialized) world.

src/client/java/minicraft/core/Renderer.java
BenCheung0422: I am not sure what you mean. It is changed from `LinkedSprite#getSheet()` into `Supplier<MinicraftImage>#get()`.

src/client/java/minicraft/core/Renderer.java
Makkkkus: How often is this lambda function called?

src/client/java/minicraft/core/Renderer.java
Makkkkus: It is a good change.

src/client/java/minicraft/core/Renderer.java
```diff
@@ -63,7 +63,10 @@ private Renderer() {
 	static float SCALE = 3;
 
 	public static Screen screen; // Creates the main screen
-	public static SpriteLinker spriteLinker; // The sprite linker for sprites
+	public static final SpriteLinker spriteLinker; // The sprite linker for sprites
+	static {
+		spriteLinker = new SpriteLinker();
+	}
```
Makkkkus, line 69: Okay, this is just plain out stupid. Just make it a static initaliser then, no need to create a static block. Functionally the same as `public static final SpriteLinker spriteLinker = new SpriteLinker();`

src/client/java/minicraft/core/Renderer.java
BenCheung0422: Actually, this lambda acts as a wrapper. Without this wrapper, `Renderer.spriteLinker.getSheet(SpriteType.Gui, "hud")` would still be required on each line rendering occuring.

src/client/java/minicraft/core/Renderer.java
BenCheung0422: Right.

src/client/java/minicraft/core/Renderer.java
Makkkkus: That doesn't answer my question.

BenCheung0422: Every time sprites within `hud.png` being rendered.

PR #519, opened 2023-06-25
Fix loadPrefsOld

BenCheung0422:
In the current `Load#loadPrefsOld`, skin index is missing and the order of language settings is not correct. Also, after my deep search, I have found that language setting is not 2.0.3-dev1, instead, the setting is valid since 2.0.4-dev3.
If my prediction is correct, loading preferences with versions 2.0.3-dev1 to 2.0.4-dev2 would throw exception `IndexOutOfBoundsException: Index 0 out of bounds for length 0`, and with versions after 2.0.4-dev2 (dev3 and later), language preference would not be loaded and for multiplayer preferences: saved IP would be saved UUID; saved UUID would be saved username; saved username would be old language preference. Or, with versions after 2.0.4-dev2, exception `ArrayIndexOutOfBoundsException: Index 1 out of bounds for length 1` might be thrown.

### DIFF
--- a/src/client/java/minicraft/saveload/Load.java
+++ b/src/client/java/minicraft/saveload/Load.java
@@ -438,49 +438,8 @@ public class Load {
 		if (prefVer.compareTo(new Version("2.0.4-dev2")) >= 0)
 			Settings.set("fps", Integer.parseInt(data.remove(0)));
 
-		SkinDisplay.releaseSkins();
-
-		// Get legacy language and convert it into the current format.
-		if (prefVer.compareTo(new Version("2.0.3-dev1")) >= 0) {
-			// Get language and convert into locale.
-			String lang;
-			switch (data.remove(0)) {
-				case "english":
-					lang = "en-us";
-					break;
-				case "french":
-					lang = "fr-fr";
-					break;
-				case "hungarian":
-					lang = "hu-hu";
-					break;
-				case "indonesia":
-					lang = "id-id";
-					break;
-				case "italiano":
-					lang = "it-it";
-					break;
-				case "norwegian":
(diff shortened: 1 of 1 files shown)


src/client/java/minicraft/saveload/Load.java
```diff
@@ -492,6 +451,48 @@ private void loadPrefsOld(String filename) {
...
+						break;
+					default:
+						lang = null;
+						break;
+				}
+
+				if (lang != null) {
+					Localization.changeLanguage(lang);
+				}
+			}
```
Makkkkus, line 494: You're aware that this block will never run this code, right?

BenCheung0422: This code might be run.

Makkkkus: It checks if the preferences version is less than 2.0.3-dev1 and more or equals to 2.0.4-dev3. That will never run.

BenCheung0422: There is an `else` with the outer `if` block, so it is greater or equals than 2.0.3-dev1.

Makkkkus: Good!

Makkkkus (approved):

ChristofferHolmesland (approved):

BenCheung0422:
An update and correction, if there are no logins in game (no related settings) with the versions before 2.0.7, as empty values are unexpectedly removed (the original design for loading does not expect values being removed, even empty), `IndexOutOfBoundsException: Index 0 out of bounds for length 0` may also occur. A fix for this: #565.

PR #523, opened 2023-06-27
Optimize CPU usage

BenCheung0422:
In my test with VisualVM profiler, it is unknown that the while loop processing without `Updater#tick()` or `Renderer#render()` invokes occupying relatively great amount of CPU time in self-time. According to my computer, this reduces averagely around 60%-70% of CPU usage in title screen and 50% in a just created world. A thread blocking/sleeping is always needed to prevent while loop looping with empty body using indefinite amount of CPU time. This also presents in the original Minicraft `Game#run()` code (with `Thread#sleep(2)`). `1.01` in the FPS time check is used for relax the hard FPS restriction to prevent from running down the maximum value.
Reference: https://stackoverflow.com/a/61598906

### DIFF
--- a/src/client/java/minicraft/core/Initializer.java
+++ b/src/client/java/minicraft/core/Initializer.java
@@ -96,7 +96,7 @@ public class Initializer extends Game {
 			}
 
 			now = System.nanoTime();
-			if (now >= lastRender + 1E9D / MAX_FPS / 1.05) {
+			if (now >= lastRender + 1E9D / MAX_FPS / 1.01) {
 				frames++;
 				lastRender = now;
 				Renderer.render();
--- a/src/client/java/minicraft/core/Initializer.java
+++ b/src/client/java/minicraft/core/Initializer.java
@@ -96,11 +96,22 @@ public class Initializer extends Game {
 			}
 
 			now = System.nanoTime();
-			if (now >= lastRender + 1E9D / MAX_FPS) {
+			if (now >= lastRender + 1E9D / MAX_FPS / 1.05) {
 				frames++;
 				lastRender = now;
 				Renderer.render();
 			}
+
+			try {
+				long curNano = System.nanoTime();
+				long untilNextTick = (long) (lastTick + nsPerTick - curNano);
+				long untilNextFrame = (long) (lastRender + 1E9D / MAX_FPS - curNano);
+				if (untilNextTick > 1E3 && untilNextFrame > 1E3) {
+					double timeToWait = Math.min(untilNextTick, untilNextFrame) / 1.2; // in nanosecond
+					//noinspection BusyWait
+					Thread.sleep((long) Math.floor(timeToWait / 1E6), (int) ((timeToWait - Math.floor(timeToWait)) % 1E6));
+				}
+			} catch (InterruptedException ignored) {}
 
 			if (System.currentTimeMillis() - lastTimer1 > 1000) { //updates every 1 second
 				long interval = System.currentTimeMillis() - lastTimer1;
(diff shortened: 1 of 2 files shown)


Makkkkus:
Doesn't this break the FPS setting?

src/client/java/minicraft/core/Initializer.java
```diff
@@ -96,12 +96,23 @@ static void run() {
...
 			}
 
+			try {
+				long curNano = System.nanoTime();
+				long untilNextTick = (long) (lastTick + nsPerTick - curNano);
+				long untilNextFrame = (long) (lastRender + 1E9D / MAX_FPS - curNano);
+				if (untilNextTick > 1E3 && untilNextFrame > 1E3) {
+					double timeToWait = Math.min(untilNextTick, untilNextFrame) / 1.2; // in nanosecond
+					//noinspection BusyWait
+					Thread.sleep((long) Math.floor(timeToWait / 1E6), (int) ((timeToWait - Math.floor(timeToWait)) % 1E6));
```
Makkkkus, line 112: You haven't really fixed the problem.
"In most cases spinning is considered an [anti-pattern](https://en.wikipedia.org/wiki/Anti-pattern) and should be avoided,[[2]](https://en.wikipedia.org/wiki/Busy_waiting#cite_note-2) as processor time that could be used to execute a different [task](https://en.wikipedia.org/wiki/Thread_(computer_science)) is instead wasted on useless activity."
- https://en.wikipedia.org/wiki/Busy_waiting

BenCheung0422: > Busy-waiting itself can be made much less wasteful by using a delay function (e.g., [sleep()](https://en.wikipedia.org/wiki/Sleep())) found in most operating systems. This puts a thread to sleep for a specified time, during which the thread will waste no CPU time. If the loop is checking something simple then it will spend most of its time asleep and will waste very little CPU time.

BenCheung0422:
> Doesn't this break the FPS setting?

I have tested, it does not break.

src/client/java/minicraft/core/Initializer.java
Makkkkus: Your IDE begs to differ. The functions might not neccecaraly be the same.

BenCheung0422: Why is it?

Makkkkus: I don't understand what you mean with "Why is it?" but, this is a line where you force the IDE to shut up.
//noinspection BusyWait

Makkkkus:
Okay, that's good.

src/client/java/minicraft/core/Initializer.java
BenCheung0422: IDE expects that using `wait()` in a loop is a worse practice because it may appears as using this function to pause the current thread *for another thread*. Here, this main thread only executes `tick()` and `render()` in desired intervals. Also, for constant interval, using timeout is better, but this does not. By the way, it would be better to separate rendering and ticking into 2 threads to use `waiting` in the better way, since it is currently waiting for both next ticking and rendering and wait for the shortest interval to execute the closest one and it would be *weird*, but concurrency and synchronization are required to be handled in an ideal way.

Makkkkus: Yeah, that sounds good.

Makkkkus (approved):

ChristofferHolmesland (approved):

BenCheung0422:
> By the way, it would be better to separate rendering and ticking into 2 threads to use waiting in the better way, since it is currently waiting for both next ticking and rendering and wait for the shortest interval to execute the closest one and it would be weird, but concurrency and synchronization are required to be handled in an ideal way.

I think it could be implemented in the near future.

PR #524, opened 2023-06-27
Add hardware acceleration setting

BenCheung0422:
By default, hardware acceleration is not enabled. With help of #523, this could further reduce around 10%-15% CPU usage (with using GPU) when hardware acceleration is enabled.


Makkkkus (changes requested):
You don't have use this as a normal setting.

BenCheung0422:
Is it because it applies only on startup?

Makkkkus:
Yeah, you could make it possible to be disabled by a launch argument.

BenCheung0422:
For convenience, it could be a setting that require a restart with modified value.

Makkkkus:
No, using a launch argument would be more convenient. Both in code and use.

BenCheung0422:
Then, a launching argument modifying setting value.

Makkkkus:
Don't use the Settings class. That will just be confusing.

BenCheung0422:
The entire setting is removed from settings and preferences. It can now only be disabled by using a flag.

Makkkkus:
You haven't fixed this yet. You need to revert the changes related to saving and remove a reference to the setting in OptionsMainMenuDisplay.

BenCheung0422:
Ok.

ChristofferHolmesland (approved):

Makkkkus (approved):

PR #525, opened 2023-06-28
Implement "curse" into the game

BenCheung0422:
There is an idea about "curse", which is the curse of "alone" applying on the player with the living Air Wizard. When Air Wizard dies, the curse is also broken, then the player can be no longer alone from animals with kind minds.
This adds behaviors to the animal cow and sheep. When the player is holding wheats with the curse broken, those animals would follow to the player with distance. Also, this fixes the problem that mob despawning too close to the player. The ability to feed animals might also be added.


BenCheung0422:
Implementation of "curse" is rejected.

PR #527, opened 2023-07-07
Fix furniture holding player sprites missing

BenCheung0422:
This reverts a change by da6d4799577222c3ad12007c7005c454a382f6ef in #413. In fact, the first line (0, 0, 64, 16) contains normal skins and the second line (0, 16, 64, 16) contains furniture holding skins, so this is a mistake of determining this as redundant. The current sheet does not include the second line, so this leads to a texture missing in furniture holding rendering.


Makkkkus:
This is a perfect example of your carelessness.

Makkkkus (approved):

ChristofferHolmesland (approved):

PR #528, opened 2023-07-07
Change screenshot scaling strategy

BenCheung0422:
Currently, bicubic interpolation is used for scaling screenshots into desired scales. ~~This adds a setting to control which interpolation is used for this.~~ (The interpolation type is entirely changed to nearest neighbor instead of adding a setting) In time complexity and file size, it is obvious that: bicubic > bilinear > nearest-neighbor. However, the quality of images differs from each other's. In usual, bicubic is commonly used for image and video processing, but the canvas of Minicraft is scaling with ratio magnification, which is nearest-neighbour. Using nearest-neighbour could make the screenshots closer to the look in-game.


Makkkkus:
Don't make it a setting, just use nearest neighbour.

Makkkkus (changes requested):
Still not fixed

BenCheung0422:
Alright, I have changed.

ChristofferHolmesland (dismissed):

src/client/resources/assets/localization/en-us.json
```diff
@@ -282,7 +282,6 @@
   "minicraft.settings.mode.hardcore": "Hardcore",
   "minicraft.settings.mode.score": "Score",
   "minicraft.settings.scoretime": "Time (Score Mode)",
-  "minicraft.settings.screenshot_scale": "Screenshot Scale",
```
Makkkkus, line 285: You removed a localisation entry?

Makkkkus (changes requested):

src/client/resources/assets/localization/en-us.json
BenCheung0422: I have removed too many lines. I have added it back now.

Makkkkus (approved):

Litorom (approved):
Cool

PR #530, opened 2023-07-08
Add Knight Statue limit and fix Obsidian Knight phase problem

BenCheung0422:
As the number of Knight Statue is unlimited, there can be many Knight Statues placed on the same destination tile. Unintended behaviours might be allowed to happen.
This also includes a fix for Obsidian Knight phase sprite problem.

### DIFF
--- a/src/client/java/minicraft/item/SummonItem.java
+++ b/src/client/java/minicraft/item/SummonItem.java
@@ -68,12 +68,15 @@ public class SummonItem extends StackableItem {
 					// If the player nears the center.
 					if (new Rectangle(level.w/2-3, level.h/2-3, 7, 7).contains(player.x >> 4, player.y >> 4)) {
 						if (!ObsidianKnight.active) {
-
-							// Pay stamina
-							if (player.payStamina(2)) {
-								level.add(new KnightStatue(5000), level.w/2, level.h/2, true);
-								Logger.tag("SummonItem").debug("Summoned new Knight Statue");
-								success = true;
+							if (level.getEntitiesOfClass(KnightStatue.class).length == 0) { // Prevent unintended behaviors
+								// Pay stamina
+								if (player.payStamina(2)) {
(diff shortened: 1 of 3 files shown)


ChristofferHolmesland:
This sounds reasonable but I'm not sure what the consequences would be.

src/client/java/minicraft/item/SummonItem.java
```diff
@@ -68,12 +68,15 @@ public boolean interactOn(Tile tile, Level level, int xt, int yt, Player player,
 					// If the player nears the center.
 					if (new Rectangle(level.w/2-3, level.h/2-3, 7, 7).contains(player.x >> 4, player.y >> 4)) {
 						if (!ObsidianKnight.active) {
-
-							// Pay stamina
-							if (player.payStamina(2)) {
-								level.add(new KnightStatue(5000), level.w/2, level.h/2, true);
-								Logger.tag("SummonItem").debug("Summoned new Knight Statue");
-								success = true;
+							if (level.getEntitiesOfClass(KnightStatue.class).length == 0) { // Prevent unintended behaviors
```
Makkkkus, line 71: This is a strange way of checking for other statues. Are you sure there isn't a better way to do this?

Makkkkus (changes requested):

src/client/java/minicraft/item/SummonItem.java
BenCheung0422: I have changed to use for loop.

Makkkkus: Yeah, then it only checks until it finds one, and not the whole array 100% of the time.

Makkkkus (approved):

Litorom:
Oh I seemed to have pushed to the wrong PR.

Litorom (approved):

ChristofferHolmesland (approved):

PR #531, opened 2023-07-08
Add knight shard

BenCheung0422:
As Obsidian Knight should drop its dedicated items as being a boss after the Air Wizard, this literally adds Knight shards (ancient and mysterious/legendary) dropped only by Obsidian Knight, and it still drops shards as being a bigger Knight. Then, more sophisticated and dedicated End Game content (materials only dropped by bosses) can be made into the game.


Makkkkus:
How many times do I have to tell you this?? I will not add any new items.

PR #436, opened 2023-02-05
Fixes a chance of deleting the world second time

BenCheung0422:
Fixes #434 
This happens because of the display handling issue. This is resolved by executor pool.

PR #535, opened 2023-07-11
Fix a problem on missing Display#onExit

BenCheung0422:
This problem causes some displays not working properly depending on `Display#onExit()` invokes. This involves changes by #436 that the display handler is using a `ArrayDeque` and loops.


BenCheung0422:
The dependency on #533 for testing purpose (with resource pack display) is removed. Furthermore, this problem does not only affect the resource pack display, but also the options displays.

ChristofferHolmesland (approved):

Makkkkus (approved):
Interresting problem, but the fix makes sense.

PR #536, opened 2023-07-12
Fix horizontal flipping of stone_wall_border.png

BenCheung0422:
This does not match the previous change in `stone_wall_border.png` because the flipping has a mistake in horizontal clipping.


Makkkkus (approved):
This is a vertical flip, but ok.

ChristofferHolmesland (approved):

PR #538, opened 2023-07-16
Tweak world generation

BenCheung0422:
- Rewrites map generation random values depending on the "distances"
  - Original `dist` calculation is irregular and concentrated near the edges.
  - Current `dist` calculation is regular, less concentrated near the edges and less deep the values.
- Irregular type of top map now varies the feature sizes depending on the map sizes.
- Sky map now varies the feature sizes depending on the map sizes.
- Top map generation tweaks:
  - Base map features:
    - Island: higher the rock amount; shallower the water sizes.
    - Box: higher the rock amount; sharper the map edges.
    - Mountain: lower the grass amount; slightly shallower the water sizes; thus, higher the rock amount.
    - Irregular: slightly higher the rock amount.
  - Lower the amount and smaller the flower features.
  - Add some small beach features with themes other than desert.
- Sky map generation tweaks:
  - Larger the cloud sizes.
- Underground map generation tweaks:
  - Dirt, water or lava regions are less concentrated at the center and some covered by rocks.
  - More sharded cave paths and smaller and fewer the open areas.
  - Slightly more ore veins but sparser the ore veins.
- Dungeon map generation tweaks:
  - Sharded lava river generates near the edges.
  - Less dirt tiles.
  - More ornate lava pools are possible generated.
  - Simple lava pools now generate, but rarely.
- Add functions to `LevelGen#main`.
- Some structures have various integrities when generating.

Waiting list of themes: flower forest and meadow.


BenCheung0422:
Also, I would like to greatly modify the way to generate (random) regional tiles. Mainly, the features of biomes (areas) would be generated first. Then, actually related tiles and biome-related structure would be generated according to the biome areas. This could make convenience when planning to have biome-restricted structure and have more well-structured world in result.

BenCheung0422:
### Proposal about New Settings and Modifications on World Generation

As mentioned in the previous comment, I would like to briefly introduce about how these changes would be implemented in the following.

**World Sizes**

I am not sure if we should tweak the options in world sizes, but if we have slightly bigger world, we could have more possibilities and contents within a world. Possibly we could have "tiny" world option instead when the sizes have been increased. The current available options are 128, 256, 512 based on the length of tiles of the square world. I would like to increase the size up to 1024 or 2048 for extensibility. Also, the design should mainly focus on the bigger world sizes instead of all sizes. There could also be special generation schemes in "tiny" worlds.

**World Types**

Currently, all available options of the world type setting are "island", "box", "mountain" and "irregular". All types consider the world edges when generating basic tiles, i.e. water/lava, dirt, grass and rock tiles, except the irregular type.
These types are only applied on the top map. Water tiles are generated only in the basic generation with types and lava tiles are only generated to replace all the placements of water tiles.

Currently, in the "island" type, water tiles are generated mainly concentrated near the edges and the number of rock tiles is reduced; in the "box" type, water tiles are forcedly generated to the edges and the shape of square is sharper; in the "mountain" type, water tiles on the edges are usually replaced by grass tiles and instead generated usually at the center, other tiles are rock tiles instead of grass tiles; in the "irregular" type, water tiles are less concentrated on the edges and more dispersed around the map. In all types, rock tiles are always centralized.

I would like to change all the options into "normal", "islands", "mountain", "plain" and "edges" ("edges" is inherited from the current option "box"). The "mountain" type keeps mainly the same in rules, but diverged from the usual rules about edges. The algorithm related to positions, negative numbers increasing near the edges exponentially in magnitudes, would be applied only the positions are near the edges. The "normal" type is mainly "irregular" from the current one, but ignoring edges in generation. The "islands" type is small islands generated on the whole map instead of a big island at the center. In "mountain" type, most of the stairs would be generated in the "mountain" and there could be "water in cave" regions. In "plain" type, there would be fewer rock tiles generated, position algorithm is ignored. In "edges" type, the edges of the square world are sharply cut and filled with all water instead, there would be less water tiles in the center Greenland, but the distributions of rock tiles would be normal.

**World Themes**

There are currently "normal", "forest", "desert", "plain" and "hell" as the options.

In sand tile generation, grass tiles are mainly replaced with sand tiles in "desert" theme and small regions of sand tiles are generated in other themes.

In tree tile generation, trees are generated discretely in groups in "plain" theme and instead more frequently in "forest" theme. The generation frequency is in between in other themes. There are also generations frequently between "forest" and other themes.

For flowers, they are generated quite frequently but in smaller scales on grass tiles. For cactuses, they are generated lonely but checked frequently on sand tiles.

For changes, I would replace the generation by biomes introduced below.

**Sky Level**

There would not be such changes on sky level, but I would like to have functions to regenerate cloud tiles and cloud ores on demand, i.e. attempting to regenerate tiles when there is lack of such tiles in the world. Also, I think there could be some structures on the cloud. In balancing, the cloud tiles should be generated less connected but in stacks instead.

**Underground Levels**

There should be more tunnel structures in the cave systems. There could also be some structures underground too. In the current structure, there is a region at the center in underground levels, which is covered by dirt in B1, water in B2 and lava in B3.
The generation of this region should be reconsidered based on balancing and structure. There could also be cave biomes.

**Dungeon Level**

The generation of the dungeon level is quite clean. There are walls on the edges and a boss room at the center. There are seldom dungeon structures on the level and might makes the abundance poor. There could be more different structures to complement this.

**Structures**

Take an overall view of the structures, it is quite lack of enough structures, including decorative and storytelling structures. For example, there could be beach, ocean, forest, underground remains.

**Biomes**

There would be (actual work might vary):
- sky level: cloud region, cloud ore region;
- top level: beach, sea, deep ocean, river, lake, pond, plain, forest, mountain, cave pond;
- underground level: cave, cave lake, cave tunnel;
- dungeon level: dungeon.

I would like to have world themes: "plain", "desert", "forest", "hell", "mountain", "normal". The generation would be generated as desired following the nature of the themes.

**Conclusion**

In conclusion, the world generation should be well-planned in practice. However, the code structure of `Level` and `LevelGen` should also be rewritten to have better structure when doing so.

PR #541, opened 2023-07-19
Move chest menu to the right

### LINKED ISSUE #537: Chest menu should be on the right
When opening a chest you would expect the chest menu to be on the right, since the inventory is normally on the left.

Move the chest menu to the right side of the screen.

**Additional context**
bilde

TheBigEye:
* The order of the inventories menus was changed, now the player's inventory will be on the left, and the chest inventory will be on the right
- This should fix #537


Makkkkus (approved):
Great!

BenCheung0422 (approved):

PR #546, opened 2023-07-28
Clean up usages of Random in chests

BenCheung0422:
When populating chests, as it involves world generation, it should also include the world generation random provider when using randoms. This fixes this problem. Unused `Inventory#tryAdd` methods are also removed. All chests should be the same for the same seed in initialization now.


Litorom:
This looks like it fixes another bug I discovered today, the dungeon chests in structures generate empty. Please fix your code and test and I’ll merge it

src/client/java/minicraft/saveload/Load.java
```diff
@@ -1129,7 +1129,7 @@ private static Entity getEntity(String string, int mobLevel) {
 			case "Workbench": return new Crafter(Crafter.Type.Workbench);
 			case "Chest": return new Chest();
 			case "DeathChest": return new DeathChest();
-			case "DungeonChest": return new DungeonChest(false);
+			case "DungeonChest": return new DungeonChest(null);
```
Makkkkus, line 1132: These will probably cause nullpointerexceptions.

BenCheung0422: `null` case is handled within `DungeonChest`.

Makkkkus: Alright, I see what you mean. But do we want the chests to be empty in those cases. And should the populate inventory function even be called from the DungeonChest class?

BenCheung0422: In `Load`, this method fetches the initial state (empty state this case) for further entity loading, although actually this is quite badly designed.

Litorom: Will empty locked chest populate?

BenCheung0422: If it is loaded from saves, the chests are never populated.

BenCheung0422: Also, the initial state of "locked" does not affect the final state of the chest. The saved value of "locked" should always be loaded from the save after the initial construction of the instance.

Litorom: Locked dungeon chests that spawn in structures method are empty while those that are spawned via the other method are populated.

BenCheung0422: image

BenCheung0422: I hope I have not misunderstood.

Litorom: I’ll do more testing, just fix the conflicts too please.

PR #548, opened 2023-08-02
Add sign

BenCheung0422:
The original pull request doing this is #366 and has been closed because of the incompatibility of the tile handling system.
Currently, there is a way do that by adding a separated save file for signs, but it still needs to be combined into a standardized system of tile data save system in the future. Signs could make more contents within structures when complementing the storyline.


ChristofferHolmesland (dismissed):

phithagod24 (commented):
Nice!

Litorom:
Could you please resolve conflicts. I believe this can work with storyline.

BenCheung0422:
Conflicts are now resolved.

Litorom (changes requested):
Sign display still shows sometimes after leaving a sign tile.

BenCheung0422:
It should be fixed now.

Litorom (changes requested):
Upon editing the sign in creative, you can use Debug cmds and switch levels when in sign edit menu

BenCheung0422:
Okay, actually that happens to every menu, even pause menu.

BenCheung0422:
In this case, #504 should have already avoided this from happening.

Litorom (changes requested):
This is not what I meant! Please change it back and add menu checks

BenCheung0422:
Okay I added it back and moved to another place.

BenCheung0422:
The code will be updated based on #555.

PR #549, opened 2023-08-02
Add reset and redo color codes

BenCheung0422:
As described in the title, this adds reset and redo color codes for text color formatting. This rewrites the code for rendering text encoded with color codes. A reset code resets the following text and a redo code sets the color before the last color code applied, and resets the color if the last code is the reset code. This also fixes some potential rendering problems of key input entries and search entries. The method does not require a leading color code now, and has the ability to use the white tint color as the overall rendering color, which can be set by the reset code.


Makkkkus:
Why is this necessary?

BenCheung0422:
For necessity, see the applications of the feature.

Litorom:
I’m assuming this is similar to Minecraft’s color codes, but can someone actually explain this to me. 
If this can make rainbow splashes I’ll merge it for that functionality. May prove useful.

PR #552, opened 2023-08-04
Separate attack action and interact action

### LINKED ISSUE #435: Improving the activation of interaction and breaking of tiles
The key of interaction and damaging with tiles is limited. There might be potential problems with the separated handlers on separated tiles with separated tools and/or items. They are standardized, on tile breaking, interacting (also handled with breaking) and with entities. Interaction should only be "interaction" (without direct damaging or breaking). Damaging and breaking should be handled as similar as the way of as entities. The interacting keys are also different from entities to tiles.

Hurt method in tiles should be handled with damage and item. Interact method in tiles should not be expected to direct damage or destroy the tiles. The keys of interact and damage/attack should be completely separated.

BenCheung0422:
Resolves #435
Many of the methods combine attack and interact when handling interactions and damage dealing, causing the combination complexity. This adds key bindings of "interact" and "inventory" to replace the original "menu" and some implementations from "attack". Originally: "attack" - `C|SPACE|ENTER`, "menu" - `X|E`' currently: "attack" - `C|SPACE`, "interact" - `X|ENTER`, `inventory` - `E`. "Inventory" now only opens player inventory menus or closes some in-game menus; "attack" now only handles direct damage and tile destroy with bare hand or any item on tiles or entities; "interact" is now for interacting with furniture, placing tile and furniture, item interactions, etc.

### DIFF
--- a/src/client/java/minicraft/core/Renderer.java
+++ b/src/client/java/minicraft/core/Renderer.java
@@ -325,19 +325,19 @@ public class Renderer extends Game {
 		}
 
 		// This renders the potions overlay
-		if (player.showpotioneffects && player.potioneffects.size() > 0) {
+		if (player.showPotionEffects && player.potioneffects.size() > 0) {
 
 			@SuppressWarnings("unchecked")
 			Map.Entry<PotionType, Integer>[] effects = player.potioneffects.entrySet().toArray(new Map.Entry[0]);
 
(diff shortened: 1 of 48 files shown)


BenCheung0422:
Update to this pull request is completed, quite a kind of major change.

BenCheung0422:
Power glove part would be further resolved by #567.

PR #554, opened 2023-08-07
Resolve some problems in the control guide

BenCheung0422:
This fixes several problems:
- The control guide does not detect inputs from controllers.
- The control guide may still detect inputs when some menus are opened.
- Duration of the movement guide (the first one): 2 seconds to 5 seconds.


ChristofferHolmesland (approved):

Makkkkus (approved):
Seems logical. I trust that you know how this works since you made it.

PR #555, opened 2023-08-09
Move torch variations into tile data

BenCheung0422:
Using duplications in the main method is not a great idea, so this changes torch tile variations into handling the tile data instead. ~~When #437 is merged, `TileModel` can be used to merge `TorchItem` into a general `TileItem` instance.~~(outdated) `TorchItem` is merged to `TileItem` with the advanced feature from `TileModel`. Tile sprite connection fields are converted to methods to suit the changes.


src/client/java/minicraft/level/Level.java
```diff
@@ -506,11 +506,15 @@ private void sortAndRender(Screen screen, List<Entity> list) {
 
 	public Tile getTile(int x, int y) {
 		if (x < 0 || y < 0 || x >= w || y >= h /* || (x + y * w) >= tiles.length*/ ) return Tiles.get("connector tile");
-		int id = tiles[x + y * w];
-		if(id < 0) id += 256;
-		return Tiles.get(id);
+		short id = tiles[x + y * w];
+		Tile tile;
+		return (tile = Tiles.get(id)) == TorchTile.DELEGATE ? TorchTile.getTorchTile(Tiles.get((short) getData(x, y))) : tile;
```
ChristofferHolmesland, line 511: This should be changed to something that is more readable. Suggestion:
```
Tile tile = Tiles.get(id);

if (tile == TorchTile.DELEGATE) {
  return TorchTile.getTorchTile(Tiles.get((short)) getData(x, y)));
}

return tile;
```

ChristofferHolmesland (changes requested):

ChristofferHolmesland (dismissed):

Makkkkus (approved):

Litorom:
Please fix conflicts.

Makkkkus (changes requested):
Actually, when looking at this again, won't this break all saves that have torches?

BenCheung0422:
The code to load torch is not removed, so it is not affected.

BenCheung0422:
Conflicts resolved.

Makkkkus:
> The code to load torch is not removed, so it is not affected.

Yes, but then it tries to load the tourch with a the old name on new saves. That will cause an error.

BenCheung0422:
I fixed that error in the latest commit recently.

PR #543, opened 2023-07-26
Add ability to horizontally scroll entries

BenCheung0422:
Fixes #486
This is a more complete version of #489.
This generally enables the ability to scroll entries horizontally and render the entries without hardly breaking or hiding the entries.
This fixes a number of issues related to non-user-friendly rendering on entries.
Features:
- Manual scrolling when the entry is longer than the menu bounds:

https://github.com/MinicraftPlus/minicraft-plus-revived/assets/74168521/30f19600-3248-4384-a8ba-0cd2683f94b2

- Improved input entry displacement handling and scrolling:

https://github.com/MinicraftPlus/minicraft-plus-revived/assets/74168521/0060b7ac-576f-4e8b-81c1-bc83da9147a1

- Static entry scrolling (scrolling text and alternating scrolling text):

https://github.com/MinicraftPlus/minicraft-plus-revived/assets/74168521/8e6d8dd0-ee25-40ff-884a-34e8397cc29f

- World list input entry fix (previous and new):

https://github.com/MinicraftPlus/minicraft-plus-revived/assets/74168521/7b3c1d26-e282-4b08-8e95-7907362749fa

https://github.com/MinicraftPlus/minicraft-plus-revived/assets/74168521/5521685d-7c81-4b3a-bd1e-1321b249f8e4

- A new common class is added: RenderingLimitingModel
Controls to entry shifting are limited to only menus internally.
Internally, entries are wrapped by MenuListEntry for the controls of entries over menu.

PR #558, opened 2023-08-20
Fix ResourcePackDisplay entry reposition problem when moving entries

BenCheung0422:
This rendering occurs when moving entries within the resource pack display menu. The rendering is because the menus are not correctly repositioned when updated.
~~Note: ~~#533~~ is required to test this.~~(outdated)


Litorom:
I can't figure out how to active a resource pack

Makkkkus (commented):
Why do you use bitwise operators here? That's a bug magnet!

BenCheung0422:
To active a resource pack, hold your shift and use you cursor on the selected pack to the right.

Litorom:
> To active a resource pack, hold your shift and use you cursor on the selected pack to the right.

You should really change that, too confusing and the instructions are off the screen. Idk how that will work with controllers.

BenCheung0422:
I thought that is intuitive enough when there is an instruction on the screen.

BenCheung0422:
However, it is the most direct way I can find to implement this.

Makkkkus:
> To active a resource pack, hold your shift and use you cursor on the selected pack to the right.

Just use the enter button in the same way you add items to a chest. Why does it have to be different??

And instructions should be on the screen.

BenCheung0422:
About tweaking this, as this is not what this pull request should be doing, later I will include such change in #581.

BenCheung0422:
And also, this is the instruction to move pack by holding shift, although it is out of screen. #543 should have solved this problem. It should not be shorten, and breaking the line is not a neat and great way.

image

Litorom:
Only fixes positioning, approved

PR #559, opened 2023-08-22
Standardize tile explosion functions by TNT and creeper

BenCheung0422:
This comes from #372. The original problem is that in explosion blacklists of TNT and creeper are differ from each other, so the lists are grouped together with `Tiles#explosionBlacklist`. Also, some missing obsidian tiles in the blacklist are added. As the blacklist is standardized, the tile explosion behaviour should also be standardized. The behaviour is standardized by a kind of tile entity, `ExplosionTileTicker`, with exploded tile replacements and removals (back to hole tiles or infinite fall tiles). Originally, creeper does not explode with exploded tiles and instead instant breaks (replacing with hole tiles and infinite fall tiles).


ChristofferHolmesland (approved):

Makkkkus:
Can't you just copy the blacklist over from the tnt to the creeper? There might be cases where we want them to be different.

BenCheung0422:
The power of explosion should be the same as many of these are made of obsidian (inconsistence in tile hardness) and there has already the difference in radius of explosion.

PR #562, opened 2023-08-22
Change numeric entity IDs into UUIDs

BenCheung0422:
For compatibility in the near future, increasing the bounds of the IDs would be necessary. A UUID can have 2^128 possible values while an integral ID can have 2^32 possible values. Also, the conversion between the string value and the instance is well-defined in UUID. Old EIDs are recorded in world load entity mapping, so that the old IDs can correspond to the entities whilst new UUIDs are generated.
If there is an account system in the future, the system would also use UUIDs for identifiers, and the player instances would use these UUIDs as their entity identifiers. Note that UUIDs should be used as the IDs for instances (especially these related to entities and players) in common for security.
It is also noticed that some entities cannot be loaded before some entities being loaded, so this problem might be potentially valid and should be resolved in the near future.

### DIFF
--- a/src/client/java/minicraft/core/Game.java
+++ b/src/client/java/minicraft/core/Game.java
@@ -4,7 +4,6 @@ import minicraft.core.io.InputHandler;
 import minicraft.core.io.Settings;
 import minicraft.core.io.Sound;
 import minicraft.entity.mob.Player;
-import minicraft.gfx.Screen;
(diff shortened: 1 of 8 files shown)


Makkkkus:
I don't see the point of this. You're increasing the data size for what purpose exactly. And what is the problem with just using an int64 if the values become too high. We're far away from that point now anyway.

PR #572, opened 2023-09-30
Added fences

GameJarne:
This adds wood, stone and obsidian fences.


ChristofferHolmesland (changes requested):
Looks good, except it's missing translations? You need to add entries for the tiles.

GameJarne:
I'll do that right now. Although en-us has more entries than the other translation files?

GameJarne (commented):
Added fences to English translation

BenCheung0422 (commented):
I think it should be allowed to place wooden fences on planks. Furthermore, I think all the fences should be allowed to be placed on grass tiles and floors with their corresponding materials.
Is it possible to add fence gates? I think it would be better if there is a door or gate that can allow players and entities passing through without breaking the tiles whilst having the similar textures with the fences.
The way to handle fence textures is clever, it is good.

src/client/java/minicraft/level/tile/Tiles.java
```diff
@@ -76,6 +76,9 @@ public static void initTileList() {
 		tiles.put((short)47, new BossWallTile());
 		tiles.put((short)48, new BossFloorTile());
 		tiles.put((short)49, new BossDoorTile());
+		tiles.put((short)50, new FenceTile(Tile.Material.Wood));
+		tiles.put((short)51, new FenceTile(Tile.Material.Stone));
+		tiles.put((short)52, new FenceTile(Tile.Material.Obsidian));
```
Makkkkus, line 81: Using there ids two times is absolutely catastrophic!

src/client/resources/resources/recipes.json
```diff
@@ -142,6 +142,32 @@
...
+            }
+          ]
+        }
+      }
+    },
+    "requirements": [
+      [
+        "has_plank"
+      ]
+    ],
```
Makkkkus, line 164: Not quite sure why this is here. I assume this will break the code.

src/client/java/minicraft/item/TileItem.java
```diff
@@ -35,16 +35,19 @@ protected static ArrayList<Item> getAllInstances() {
 		items.add(new TileItem("Plank", new LinkedSprite(SpriteType.Item, "plank"), new TileModel("Wood Planks"), "hole", "water", "cloud"));
 		items.add(new TileItem("Plank Wall", new LinkedSprite(SpriteType.Item, "plank_wall"), new TileModel("Wood Wall"), "Wood Planks"));
 		items.add(new TileItem("Wood Door", new LinkedSprite(SpriteType.Item, "wood_door"), new TileModel("Wood Door"), "Wood Planks"));
+		items.add(new TileItem("Wood Fence", new LinkedSprite(SpriteType.Item, "wood_fence"), new TileModel("Wood Fence"), "grass"));
```
Makkkkus, line 38: What about planks, sand, etc?

src/client/java/minicraft/level/tile/Tiles.java
Litorom: You are aware that this is because that it was made using another outdated version of the code therefore it’s technically correct at that time. However now it has to be adjusted

Makkkkus: The point of a merge is to adjust the written code to reflect changes upstream. It wouldn't be hard to fix.

src/client/resources/resources/recipes.json
Litorom: It shouldn't break the code as it follows the format

PR #573, opened 2023-10-08
Fix mobs spawning

BenCheung0422:
This fixes that mobs cannot be normally spawned because of the misconfigurations. The main reason is that the values are set as the tile distances for entity despawn distances. A use of using `Math#hypot` is prevented to speed up the calculation.


ChristofferHolmesland (approved):

Makkkkus (approved):
This will be revised when we add infinte terrain.

PR #579, opened 2023-10-30
Further optimize rendering methods

BenCheung0422:
This change performs well together with #523 and #524.
As a minor change, there is no longer an error popup when there is any explicitly or implicitly incompatible images regarding their sizes. Instead, a warning message is rendered in the terminal.
Due to the change to the compatibility of the rendering of the on-screen keyboard, the textures of the special keys are updated and saved as assets instead of being constructed on demand in runtime. It is mainly because the visibility of `Screen#pixels` is decreased,
There is no longer a light screen constructed with `Screen`. Instead, an object representing the overlay is present in the screen object.
Other minor changes:
- `Screen` is constructed with a parameter of `BufferedImage` instead of using the initialization method `#init(int[])` to initialize the pixel array.
- `Screen` now stores `BufferedImage` for the ability to render through `Graphics2D` on the image object.
- Most of the rendering methods except sprite rendering (compatibility problem) are replaced by using `Graphics2D` provided method.
- Methods to render an arbitrary size of area of `MinicraftImage` are added, because of the change to the on-screen keyboard.
- All methods to render now add queues instead of being rendered instantaneously. All the renderings are executed only when `Screen#flush()` is called and all processes before the latest screen clean rendering are eliminated.
- The drop in FPS on the world surface is significantly diminished.

Some methods may be still using array looping to render.

As a result, the CPU run time has been improved.
The below comparison is tested based on my personal computer.
Comparing with the current implementation (with #523):
- Before entering a world: improved by around 50%
- In world: improved by around 30%
Comparing with #524 added (comparing without this pull request):
- Before entering a world: improved by around 60%
- In world: no obvious change


Litorom:
Fix Conflicts and I’ll merge

PR #584, opened 2023-11-14
Push localizations from POEditor

BenCheung0422:
This would be the final push from POEditor once the migration is confirmed. When I have locked the localizations, this will be confirmed, and the localizations will be pulled into a new platform, Lokalise. This means that we will have a new platform to replace POEditor. This will be explained later with a new pull request with Lokalise replacing the role of POEditor in this project.

POEditor Sync is temporarily used for synchronization between this repository and POEditor. The configuration file will be removed once the project on POEditor is locked.


Litorom (approved):
Approved; no code changes, only updated translations.

El-Virus:
We'd need people to review the translations, some strings aren't translated properly.

Litorom:
> We'd need people to review the translations, some strings aren't translated properly.

Umm, I'm pretty sure these were reviewed by the site we were using. If some translations aren't correct, I still think this should be merged and fixed later because we need the correct translations

ChristofferHolmesland (approved):

PR #586, opened 2023-11-15
Improve key detection support in InputHandler

BenCheung0422:
This is a remake of both #515 and #533. This resolves the input key detecting problems in both modifiers and menu input transitions. An abstract class is added in this case, also ensuring safety for keys, without directly modifying the fields or states of the keys. No support for compounded inputs is removed.
For now, `+` is necessary for compounded inputs consisting of key mappings. The problems (related bugs) some menus and actions not detecting properly have been resolved.


Litorom (approved):
Works fine with my tests! Search menu bug is fixed

El-Virus (approved):

PR #588, opened 2023-11-19
Update controller code for key detection and add setting for controller enablement

BenCheung0422:
*This depends on #587*
A setting for controller enablement is added, with a default value of disabled. It detects controller configurations now if and only if it is enabled. It now automatically searches for another controller plugged (and selects the first result) when the using one is no longer plugged. It is currently impossible to provide an ability to specify a controller connected because of the limitations of the currently using library.
The code for key status is also updated, and it is now using the similar mechanism to keyboard key input detections.


BenCheung0422:
New status information for both controller availability and input method, added with the help of the addition of the status bar from #587.

https://github.com/MinicraftPlus/minicraft-plus-revived/assets/74168521/eb1bfd81-c044-4052-807b-50d60a5c08bd

PR #592, opened 2023-12-27
Added Boat

Litorom:
The ultimate and so useful feature has been added. Introducing the boat, the float on water that consumes no stamina!

All jokes aside: 

Additions:
- Boat Entity
- Boat Item

Notes:
- Breakable via any tool item
- Boats are faster than the player on water to add a benefit and reason to use them


El-Virus:
Ah, CodeQL reminds you that the input system has been changed, do consider syncing your branch to solve this.

src/client/java/minicraft/entity/rideable/Boat.java
El-Virus, line 1: Quick question, why do you sometimes use the inline ifs but in some cases you add braces?

El-Virus (changes requested):

El-Virus:
Right, you'll also have to modify your code to use the new input system.

src/client/java/minicraft/entity/rideable/RideableEntity.java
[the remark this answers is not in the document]
Litorom: This is IntelliJ’s autocomplete. Burn should always be true, unless like obsidian boats. Swim should always be true for aquatic creatures or vehicles.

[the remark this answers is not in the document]
El-Virus: In that case, canSwim should be changed to `return true` and canBurn could be removed

src/client/java/minicraft/item/Recipes.java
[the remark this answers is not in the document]
Litorom: Actually I’m stupid and forgot to add a craftable recipe for boats. With that I’ll fix the line endings

src/client/java/minicraft/entity/rideable/RideableEntity.java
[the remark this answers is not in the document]
Litorom: Okay

BenCheung0422:
What about following the rule of "everything costs stamina"? Using boat on water still costs stamina, but it consumes stamina per a certain distance travelled continuously.

src/client/java/minicraft/item/FurnitureItem.java
```diff
@@ -108,6 +141,9 @@ public boolean isDepleted() {
 	public @NotNull FurnitureItem copy() {
 		// in case the item is a spawner, it will use the sprite position (sx, sy)
 		// instead if it is not, the constructor will obtain said sprite
-		return new FurnitureItem(furniture.copy());
+		if (!isRideable)
+			return new FurnitureItem(furniture.copy());
+		else
+			return new FurnitureItem(rideable);
```
Makkkkus, line 147: .copy()

src/client/java/minicraft/entity/rideable/RideableEntity.java
```diff
@@ -0,0 +1,45 @@
+package minicraft.entity.rideable;
+
+import minicraft.core.Game;
+import minicraft.entity.Entity;
+import minicraft.entity.mob.Mob;
+import minicraft.entity.mob.Player;
+import minicraft.gfx.Screen;
+import minicraft.gfx.SpriteLinker;
+
+public class RideableEntity extends Entity {
```
Makkkkus, line 10: Why does this class exist? It does nothing different from `Boat`.

src/client/java/minicraft/entity/rideable/Boat.java
```diff
@@ -0,0 +1,292 @@
...
+            int yt = y >> 4; // the y tile coordinate that the entity is standing on.
+
+            // Calls the steppedOn() method in a tile's class. (used for tiles like sand (footprints) or lava (burning))
+            level.getTile(xt, yt).steppedOn(level, xt, yt, this);
+        }
+
+        return !stopped;
+    }
+
+    public boolean move2(double xa, double ya) {
```
Makkkkus, line 173: `move` and `move2` should probably use the corresponding methods in the `Entity` class.

src/client/java/minicraft/entity/rideable/Boat.java
```diff
@@ -0,0 +1,292 @@
...
+			}
+		}
+	}
+
+
+    @Override
+    public Boat clone() {
+        return new Boat();
+    }
+}
```
Makkkkus, line 292: This class must probably override the `die()`  method to make sure dieing works as intentional.

src/client/java/minicraft/entity/rideable/Boat.java
```diff
@@ -0,0 +1,292 @@
...
+	private static final SpriteLinker.LinkedSprite itemSprite = new SpriteLinker.LinkedSprite(SpriteLinker.SpriteType.Item, "boat");
+	public Player playerInBoat = null;
+
+	private int exitTimer = 0;
+	private boolean boatAnim = false;
+	protected int pushTime = 0;
+
+	private int tickTime = 0;
+
+	private Direction pushDir = Direction.NONE; // the direction to push the furniture
```
Makkkkus, line 36: Check if the fields above are already defined in the `Entity` class.

Makkkkus (commented):
My main impression from this is marked by a feeling of confusion, but that is not your fault I suppose.

I don't know if I fully addressed your concerns, but the following is what I managed to find.

src/client/java/minicraft/item/FurnitureItem.java
Litorom: There is no real copy class for rideable

src/client/java/minicraft/entity/rideable/Boat.java
Litorom: I’ll trt

BenCheung0422:
For people interested, this is my version of implementation for boat: https://github.com/BenCheung0422/minicraft-plus-revived/tree/boat

BenCheung0422:
It is a question to whether to consume stamina (by travelling ticks). Anyway, Litorom has merged my branch to this pull request and I have also made some further changes.

PR #595, opened 2023-12-29
Publish gradle build to github packages

Makkkkus:


ChristofferHolmesland (dismissed):

Litorom:
To be honest, idk what this does

BenCheung0422 (dismissed):
Do we have these secrets? It seems like there is no obvious problem with the configs, so I give it an approval.

PR #601, opened 2023-12-30
Fix steppedOn and bumpedOn x and y positions

KalmeMarq:
In Entity::moveByEntityHitBoxChecks, when calling the bumpingHandler and steppingHandler the horTile and frontTile were switched. This fixes sand particles spawning in the wrong tiles.


BenCheung0422 (approved):
Good spot. I think it resolves.

Litorom (approved):
I actually forgot to approve when I tested this. I saw no problems and appears to have fixed the bug.

PR #603, opened 2024-01-04
Resolve any potential NullPointerException throws caused by Entity#getClosestPlayer()

### LINKED ISSUE #602: When the Air Wizard kills you, the game throws NPE.
**To Reproduce**

1. Spawn Air Wizard
2. Die
3. See error

**Expected behavior**
Not NPE...

**Screenshots**
image

**Desktop:**
- OS: Linux
- Game version: 2.2.0-dev4

**Additional context**
I initially thought something I was doing broke it, until I recreated it in vanilla.

BenCheung0422:
Resolves #602
A `NullPointerException` would be thrown as there are missing null handles.
This kind of potential bugs would be reduced by the time #512 is finished.

Known ways to reproduce:
- Player dies when there is a spark exists (can be generated by AirWizard) (? to 2.2.0-dev4)
- Player dies when Obsidian Knight is active. (Unchecked) (2.2.0-dev3 to 2.2.0-dev4)

### DIFF
--- a/src/client/java/minicraft/entity/Entity.java
+++ b/src/client/java/minicraft/entity/Entity.java
@@ -376,6 +376,7 @@ public abstract class Entity implements Tickable {
 	 *
 	 * @return the closest player.
 	 */
+	@Nullable
 	protected Player getClosestPlayer() {
 		return getClosestPlayer(true);
 	}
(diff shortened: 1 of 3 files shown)


Litorom (approved):
I don’t need to test this. I see it works because it adds more checks and should work as intended.

ChristofferHolmesland (approved):

PR #607, opened 2024-01-06
Reformat codes with some spacing

BenCheung0422:
Added spaces before array initialization, within array initialization, code braces.
I think single line code braces should not be reformatted if they are short to reduce line uses.
A line feed after description in JavaDoc is removed to reduce line uses.
Smart tabs is enabled.
More as described in EditorConfig file if there is no missing config.
Simple one-line codes should not be reformatted to reduce line uses and ensure some neatness.
JavaDoc param and throw description is no longer aligned for convenience.


Makkkkus (approved):
This looks good.

Litorom:
Please fix conflicts

Makkkkus (approved):

PR #608, opened 2024-01-07
Fixed Tnt interaction

TheBigEye:
_This bug seems to have been present for too long_

- Previously, when a player tried to pick up the placed Tnt to put it in their inventory, they could not do so and it would activate and explode accidentally, this because we did not handle that case within the appropriate method

- The player can now collect TNT again as long as it has not been previously activated


BenCheung0422 (approved):
From the file changes, it seems like correct.

Litorom (approved):

PR #610, opened 2024-01-09
File backups in development builds

BenCheung0422:
There has already been an incident that there is a break of preferences by 9de1e2a200cec588d80e21d9a2bd4b3df976104d (originated from #425; attempt of fix as in #503). Ideally, this kind of save file scheme changes should always be conducted and followed-up so as to preserve the quality of version continuation on save files, even on pre-releases. This is related to the quality of pre-releases.

Then, in case of save file corruptions, this allows backups of both preferences and world saves. Backups of preferences are performed automatically when saving and upgrading preferences with development builds; backup prompts are shown when users are attempting to load old worlds in development builds. (This might be extended to generally all versions, and all loadings of conflicting versions respectively.)

(Edited on 2023/01/11)
<details><summary>Old summary</summary>
<p>

> Recently, due to the changes to the workflows of development builds, pre-releases, the risks of getting corrupted files may significantly increase. This allows backups of both preferences and world saves. Backups of preferences are performed automatically when saving preferences with development builds; backup prompts are shown when users are attempting to load old worlds in development builds.

</p>
</details>


src/client/java/minicraft/saveload/Version.java
```diff
@@ -82,6 +82,11 @@ public int compareTo(@NotNull Version ov, boolean ignoreDev) {
 		return 0; // The versions are equal.
 	}
 
+	public int getMake() { return make; }
+	public int getMajor() { return major; }
+	public int getMinor() { return minor; }
+	public int getDev() { return dev; }
```
Makkkkus, line 88: Remove these

Makkkkus (dismissed):

src/client/java/minicraft/saveload/Version.java
BenCheung0422: Ok

PR #611, opened 2024-01-10
Rate limit for autobuild.yml

BenCheung0422:
Now this workflow should now be able to run only when there is no successful or running auto build job since the latest commit not committed by Actions. This runs every time a pull request closed, manually or every 4 hours. A check is done before running. Within the check, it first fetches the commit datetime of the latest commit not committed by Actions (number of items: 30); then it searches for the presence of any successful or running or pending workflow runs (number of items: 30); it sends the boolean value to the environment about to continue the job or not, it is true if the presence is false; if the sent value is false, it aborts (cancels the workflow run). The build job requires conditions to run first, so that it will run only when conditions are successful.

### DIFF
--- a/.github/workflows/autobuild.yml
+++ b/.github/workflows/autobuild.yml
@@ -1,11 +1,34 @@
 name: Nightly build
 on:
+  pull_request:
+    types: [closed]
+    branches: ["main"]
   schedule:
     - cron:  '0 0 * * *'
   workflow_dispatch:
   
 jobs:
+  conditions:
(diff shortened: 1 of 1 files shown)


Litorom:
Tbh idk how to test this without merging. I’ll revert the merge if it doesn’t work

BenCheung0422:
After looking on the GitHub docs, actually `shell: bash` is needed for the configs. I was assuming the commands were running with a bash shell. Anyway, this part is now not needed anymore.

PR #581, opened 2023-11-06
Adjust and improve interface structures and GUI experience

BenCheung0422:
*This depends on #543, #610, #618, #500 and #549.*
*Related issue: #633*
This replans all confusing displays, many menu structures, reviews all localizations, and minorly includes changes related to menus and localizations. Some standardization and generalization will be included. Such parts will be included here later.

Changes to world creation menu:
- The size of the menu is enlarged.
- Default world name: "New World".
- "Create World" entry is fixed as the last entry.
- "Enter World Name" entry is renamed to "World Name" and fixed as the first entry.
- "Trouble with world name?" entry is removed.
- The remaining entries are arranged as (up-to-down): "Seed", "World Size", "Theme", "Terrain Type", "Quests", "Tutorials", "Game Mode".
- Some minor options are renamed.
- "Create World" button is now disabled when the world name is not valid, as in the design in the repair bench.
- All entries will be aligned centered.
- Input entries are now rendered in white when selected and valid.
- The input will now be resolved when there is conflicts of the world name input.
  - For duplication, the name will be incremented in format of "<name> (<number>)" as Windows does.
  - World name preview is placed under the input entry.

Changes to the "troublesome input" entry:
- The entry is removed from world generation display.
Instead, a warning popup will be displayed in the key binding settings once the user set the keys. When the user first set the keys, an information popup will be shown, also providing a prompt hotkey for a revert. When exiting the menu, a warning confirmation popup will be shown.

Addition of "How to Play" menu:
- Contents is integrated from `instructions.txt` book/text and "Controls" instructive display menu.
- The entry is placed under help menu from the title display with the old position of "instructions" menu.
- "Controls" entry is removed from options.
- Input instructions from menus are removed, and mostly integrated into the "How to Play" menu
- There is now only a line of text "(Only keyboard input is supported)" to replace the instructions on the bottom of the title display.
- There are new rules regarding the input instructions.

Changes to world selection menu:
- There are gaps between the title, world list menu and the description box.
- When the user is trying to loading a world (and world version is not the "current" pre-release) from a pre-release, a prompt will be shown. An option to backup the world will be provided.
- The world list will be arranged by the time last entered from the latest to the oldest.
- The world list will be placed just under the title. There can be at most 6 to 8 worlds shown at once in the list.
- The world version ("Version"), mode, the time last played will be shown in a menu box for the world selected by the cursor under the list.
[...]

PR #618, opened 2024-02-01
Separate HistoricLoad from LegacyLoad

BenCheung0422:
*This depends on #610.*
It is well-known and obvious that both `Save` and `Load`, even `LegacyLoad` are messy and chaotic. I separate the part of version loading that is not indicated by versioning. With my investigation, world save versioning is added in 1.9.1, but save system is implemented earlier, which means there are certain versions that there are no version indicators. Then, I reorganize to check the history and to clearly write a new loading mechanism for it. Ideally, this method could be beneficial for all loading behaviour, but it is a kind of more works to do, and I do this is just for the part of #581.
A on-demand auto data fixer is included.


Makkkkus (commented):
I suggest not touching LegacyLoad. It works as is, and is a collection of everything using the old save feature. That is fine, and should in theory just work indefinitly (if it isn't touched). I therefore advice that this pull request should be closed.

BenCheung0422:
There should be no big problem when the codes are not in a mess.

BenCheung0422:
Also, can you sure it can 100% work while it is not changed? It needs to be tidied up.

BenCheung0422:
Now when more problems fix, it is now working. Revisions for initial worlds tested: 9a09f539, 862a2e58, 585fef3

BenCheung0422:
HistoricLoad should now be completed.

Makkkkus:
I am still sceptical about changing this. LegacyLoad is fine as is all functions in it uses the old save format, and the current load file uses the new save format.

BenCheung0422:
Actually no, there were bugs with the old LegacyLoad class after tests. One of them is the `IndexOutOfBoundsException` with the rare corrupted worlds. Also, the current design of world loading classes and methods is not really safe and can easily crash.

PR #621, opened 2024-02-10
Complement information and handles regarding Unix distributions

### LINKED ISSUE #614: Gamepad loading errors on some Unix distributions
The following errors occur when I have installed such Unix distributions (amd64) with mostly default configurations. Tests are conducted with the latest 2.2.0-dev5. Screenshots are used because it is difficult to capture the texts directly.

- CentOS 7 (Core):
image
image
Obviously there is something wrong with the SDL2 library in the system.
- FreeBSD 14 (14.0-RELEASE-p4):
image
It seems like the Java library cannot correctly detect SDL2 libraries. Note that there was a similar problem in JNA but solved afterwards: https://github.com/java-native-access/jna/issues/302.

Tested distributions without this problem:
- Ubuntu 22.04
- Debian 11/12
- Fedora Workstation 39 (bundled JDK is missing AWT; a separated installation of JRE is required)

Distributions to be tested:
- ArchLinux
- Linux Mint
- OpenSUSE
- Manjaro
- Elementary
- Zorin

Follow-up actions are required for this bug. Probably some distributions require special countermeasures.
This report will be updated for any later changes.

BenCheung0422:
Resolves #614
Some words are added in README to tell the supports.
Now, errors found in #614 are caught in gamepad initialization. If the initialization failed, the feature is disabled.


Litorom (approved):
While, I can’t test this properly since I don’t have the ability to use a different OS. By looking at the code just adds more checks to insure protection against unsupported OS code.

Makkkkus (approved):
This is okay. In the future GLEW would be more appropriate for handling input.

PR #623, opened 2024-02-10
Turn sand dust into sand footsteps

BenCheung0422:
Originally the texture of this is based on tile textures, but after, it is changed into sand dusts due to changes to the textures. Recently, some players suggested turning dusts back to footsteps for appearance. This performs the change.
The particles are now generated periodically instead of randomly, so they can be aligned in a certain order.


Litorom:
Idk what you did but I can't make a new world with this patch. It gets stuck at Level 0.
image
Seed: 02122024

PR #631, opened 2024-02-16
Tweak iron and gold recipe cost down to 3 ores

BenCheung0422:
There is a view that both iron and gold are quite expensive if it costs 4 ores, so they are tweaked down to 3 ores.


Litorom:
This isn’t hard to do, anyone can change the value

BenCheung0422:
Yes, it is true.

PR #635, opened 2024-02-25
Fix book item NullPointerException

BenCheung0422:
A `NullPointerException` occurred when interacted because the book content of the default book item has been set to `null`. The change was made by the resource pack update but later not updated correspondingly. This fixes this issue and set the book content back to the default empty one.


Litorom:
Sorry wrong button

PR #636, opened 2024-02-25
Add Tall Grass

BenCheung0422:
This adds tall grasses. There are several behaviours related to tall grasses, including natural generation and wheat seeds droping when broken.
This might work better with flowers added in #445.
Changes:
- World generation:
  - Tall grass tiles are generated together with flowers.
  - Flower tiles have slightly some chance to be generated in different variants in the same generation vein.
  - Generation density of generation vein now varies randomly.
  - Size of generation vein is slightly larger.
  - Chance to generate flower is smaller than tall grass
- Grass tile:
  - Small chance to generate tall grass each tick.
  - Chance to get wheat seeds is slightly smaller.
  - Fertilizer can be applied to forcedly generate tall grass, with a chance of 1/3.


BenCheung0422:
This *may* be postponed to 2.3 because of the textures.

PR #637, opened 2024-03-11
2.5D Wall Textures

TheBigEye:
### Summary
+ This fixes style inconsistencies about wall textures, as they should be projected in a 3D perspective just like certain tiles like rocks or clouds do, as discussed [here](https://discord.com/channels/280723930942013440/1208127353599959130) ...

### A small preview
image


Litorom (approved):

BenCheung0422 (approved):
The change to have these wall textures consistent in views looks good, but still, without the help of tile (texture) models, they still look compressed, compared to the ones in RimWorld. This is an improvement to the textures.

PR #574, opened 2023-10-10
Add recipe advancement adder to Recipes.java

BenCheung0422:
This provides an ability to generate `recipes.json` automatically with minor manual conflict fixing.
This is directly copied from #445 because this occupies a high portion of changes within that pull request. This is regenerated again once either this or that pull request has been merged.

Recipe advancements removed (by keys):
- minicraft.advancements.recipes.rock_sword (Rename)
- minicraft.advancements.recipes.rock_axe (Rename)
- minicraft.advancements.recipes.rock_hoe (Rename)
- minicraft.advancements.recipes.rock_pickaxe (Rename)
- minicraft.advancements.recipes.rock_shovel (Rename)
- minicraft.advancements.recipes.rock_bow (Rename)

Recipe advancements added (by keys):
- minicraft.advancements.recipes.stone_sword (Rename)
- minicraft.advancements.recipes.stone_bow (Rename)
- minicraft.advancements.recipes.stone_pickaxe (Rename)
- minicraft.advancements.recipes.watering_can
- minicraft.advancements.recipes.stone_axe (Rename)
- minicraft.advancements.recipes.arcane_fertilizer
- minicraft.advancements.recipes.stone_shovel (Rename)
- minicraft.advancements.recipes.stone_hoe (Rename)
- minicraft.advancements.recipes.obsidian_poppet

PR #640, opened 2024-03-15
Fixed unlocked iron and gold recipes not matching actual recipe

zandgall:
Right now, players do not have access to the correct iron and gold bar recipes in the furnace. Updated "resources/recipes.json" to reflect new furnace recipes.


BenCheung0422:
Note that #574 can deal with all the changes to recipes.

PR #668, opened 2024-04-30
Lokalise: Translations update

Litorom:
This pull request was initiated by Lokalise (user Ben Cheung) at 2024-04-30 16:02:39


Litorom:
This is weirdly done, why does it use my account without a bot message?

BenCheung0422:
I think it is because Lokalise uses the user's token to create pull request instead of using GitHub app.

BenCheung0422:
Handled by #676

PR #684, opened 2024-06-07
Add copyright header

BenCheung0422:
After investigation of GPL license, it requires a copyright notice included in each project file [GNU.org](https://www.gnu.org/licenses/gpl-howto.html) [StackOverflow answer](https://softwareengineering.stackexchange.com/a/357796).
The follow header is added to each Java source files:
```java
/*
 * Copyright (c) 2024 Minicraft+ Developers and Contributors
 * SPDX-License-Identifier: GPL-3.0-only
 */
```

### DIFF
--- a/src/client/java/minicraft/core/Action.java
+++ b/src/client/java/minicraft/core/Action.java
@@ -1,3 +1,8 @@
+/*
+ * Copyright (c) 2024 Minicraft+ Developers and Contributors
+ * SPDX-License-Identifier: GPL-3.0-only
+ */
+
 package minicraft.core;
 
 @FunctionalInterface
--- a/src/client/java/minicraft/core/Condition.java
+++ b/src/client/java/minicraft/core/Condition.java
@@ -1,3 +1,8 @@
+/*
+ * Copyright (c) 2024 Minicraft+ Developers and Contributors
+ * SPDX-License-Identifier: GPL-3.0-only
+ */
+
 package minicraft.core;
 
 @FunctionalInterface
--- a/src/client/java/minicraft/core/CrashHandler.java
+++ b/src/client/java/minicraft/core/CrashHandler.java
@@ -1,3 +1,8 @@
+/*
+ * Copyright (c) 2024 Minicraft+ Developers and Contributors
+ * SPDX-License-Identifier: GPL-3.0-only
+ */
+
 package minicraft.core;
 
 import kong.unirest.Empty;
--- a/src/client/java/minicraft/core/Game.java
+++ b/src/client/java/minicraft/core/Game.java
@@ -1,3 +1,8 @@
+/*
+ * Copyright (c) 2024 Minicraft+ Developers and Contributors
+ * SPDX-License-Identifier: GPL-3.0-only
+ */
+
 package minicraft.core;
 
 import minicraft.core.io.InputHandler;
--- a/src/client/java/minicraft/core/Initializer.java
+++ b/src/client/java/minicraft/core/Initializer.java
@@ -1,3 +1,8 @@
+/*
+ * Copyright (c) 2024 Minicraft+ Developers and Contributors
+ * SPDX-License-Identifier: GPL-3.0-only
+ */
+
 package minicraft.core;
 
 import minicraft.core.io.FileHandler;
--- a/src/client/java/minicraft/core/MonoCondition.java
+++ b/src/client/java/minicraft/core/MonoCondition.java
@@ -1,3 +1,8 @@
+/*
+ * Copyright (c) 2024 Minicraft+ Developers and Contributors
+ * SPDX-License-Identifier: GPL-3.0-only
+ */
+
 package minicraft.core;
 
 @FunctionalInterface
--- a/src/client/java/minicraft/core/Renderer.java
+++ b/src/client/java/minicraft/core/Renderer.java
@@ -1,3 +1,8 @@
+/*
+ * Copyright (c) 2024 Minicraft+ Developers and Contributors
+ * SPDX-License-Identifier: GPL-3.0-only
+ */
+
 package minicraft.core;
 
 import minicraft.core.CrashHandler.ErrorInfo;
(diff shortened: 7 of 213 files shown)


Makkkkus:
You need to actually add the GPL license to the header. The proposed header indicates that the project has all rights reserved.

BenCheung0422:
The SPDX identifier should be able to do this.

BenCheung0422:
Updated the header:
```
/*
 * SPDX-FileCopyrightText: 2011-2024 Minicraft+ Developers and Contributors
 * SPDX-License-Identifier: GPL-3.0-only
 */
```

Makkkkus:
As per the license:
>How to Apply These Terms to Your New Programs
>
>If you develop a new program, and you want it to be of the greatest
>possible use to the public, the best way to achieve this is to make it
>free software which everyone can redistribute and change under these terms.
> To do so, attach the following notices to the program.  It is safest
> to attach them to the start of each source file to most effectively
> state the exclusion of warranty; and each file should have at least
> the "copyright" line and a pointer to where the full notice is found.
> 
> {one line to give the program's name and a brief idea of what it does.}
> Copyright (C) {year}  {name of author}
> 
> This program is free software: you can redistribute it and/or modify
> it under the terms of the GNU General Public License as published by
> the Free Software Foundation, either version 3 of the License, or
> (at your option) any later version.
> 
> This program is distributed in the hope that it will be useful,
> but WITHOUT ANY WARRANTY; without even the implied warranty of
> MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
> GNU General Public License for more details.
> 
> You should have received a copy of the GNU General Public License
> along with this program.  If not, see <http://www.gnu.org/licenses/>.

(https://github.com/MinicraftPlus/minicraft-plus-revived?tab=GPL-3.0-1-ov-file#readme)

So it would be better to follow the convention and add the following as a header:

    /*
     * Minicraft+ is a 2D survival sandbox game.
     * Copyright (C) 2011-2024 Minicraft+ contributors
     *
     * This program is free software: you can redistribute it and/or modify
     * it under the terms of the GNU General Public License as published by
     * the Free Software Foundation, either version 3 of the License, or any
     * later version.

     * This program is distributed in the hope that it will be useful,
     * but WITHOUT ANY WARRANTY; without even the implied warranty of
     * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
     * GNU General Public License for more details.

     * You should have received a copy of the GNU General Public License
     * along with this program.  If not, see <http://www.gnu.org/licenses/>.
     */

BenCheung0422:
> It is safest to attach them to the start of each source file to most effectively
state the exclusion of warranty; and each file should have at least
the "copyright" line and a pointer to where the full notice is found.

Here, it mentioned that *at least* the copyright line and a pointer, where it should be sufficient with SDPX header. As explained by someone:
> The legal language of GPL3 does not refer to 'machine readable' or 'plain readable'. In order to accommodate both (and to comply with the license terms) you should include the machine readable SPDX into every single file and in addition have a license.md with the complete language of the license, the copyright and other attribution notices.

Source: https://opensource.stackexchange.com/a/12047

It means that SDPX headers here would be enough to identify it as licensed under GPLv3.

Makkkkus:
I guess that isn't totally unreasonable.

I propose changing the copyright text to `Copyright (C) 2011-2024 Minicraft+ contributors`, as that is how most other projects does it. See https://www.openstreetmap.org/copyright/en.

In addition, one must be certain that all files are created by members of the project and ensure that each file contains no blocks of original source code. Although, permission from Notch on this specific issue is should be enough to relicense all files.

BenCheung0422:
Okay, I have updated the header as:
```java
/*
 * SPDX-FileCopyrightText: 2011-2024 Minicraft+ contributors
 * SPDX-License-Identifier: GPL-3.0-only
 */
```
And I have compared the current code with the original code ([reference source](https://github.com/skeeto/Minicraft)), and summarized as below:
| Original Source Method/Class | Current Minicraft+ Method/Class | Remarks |
| -------------------------------- | ----------------------- | -------------------------- |
| `Game#renderFocusNagger()` | `Renderer#renderFocusNagger()` | Mostly more or less unchanged, but more comments added | `LevelTransitionDisplay` | `LevelTransitionDisplay` | Mostly the same, but the class is small |
| `LevelGen` | `LevelGen` | Except some lines of additions and changes due to systems, mostly the same as the original |
| Overrides of `Tile#interact()` | Overrides of `Tile#interact()` | Mostly the same in logic |
| `InfiniteFallTile` | `InfiniteFallTile` | Mostly the same, but the class is small |
| `Level#renderBackground()` | `Level#renderBackground()` | Mostly the same, but with changes to values and the size is not big |
| `ItemEntity#tick()` | `ItemEntity#tick()` | Mostly the same, but more comments added |
| `Mob#findStartPos()` | `MobAi#checkStartPos()` | Logically similar, but with some different code |
| `TextParticle#tick()` | `TextParticle#tick()` | Mostly the same, but more comments added |

Additional remarks:
- Some menus contain mainly just the transformations of implementation from the original source because of the entry system.
- Methods contained only or less than 5 simple code lines are not counted in.
- Sometimes, just more additions of code are added with respect to the original code in some methods, but not counted in.

In addition, other than the codebase, another concern is about the assets (which may not be licensed under any media licenses), including sprites and audio, may have been copyrighted, but still under the similar concern of licensing as like the original codebase.

Makkkkus:
Great work looking into this!

Although I believe that files with original code should have a header that explains the situation, saying that "specific portions of this file (functions) are not licensed as GPL-3".

We'll have to look into audio and such on another occation.

BenCheung0422:
Actually, I think the situation should be explained in the root license file, or otherwise in another file in the root.

Makkkkus (approved):

Makkkkus:
I think this should probably be implemented at some point.

Given that this is already so out of date I think this should be closed. The idea is good nonetheless, and the disucssion is very valuable.

Issue #685, opened 2024-06-18
Dropped items don't pick up. And Z menu bugs.


hilcon222:
**Describe the bug**
Sometimes, when I open personal crafting menu with Z key, instead of crafting the items, they randomly drop on the floor and I can't pick them up.

**To Reproduce**

1. Click Z to open personal crafting
2. Craft a wood plank
3. It consumes wood but you see that you still have 0 planks
4. Close crafting menu
5. Look on the ground to see a lotta planks you can't pick up.

**Expected behavior**
I wanted planks in my inventory without them dropping.
And if i drop an item, im supposed to be able to pick it before it despawns.

**Desktop (please complete the following information):**

- OS: Windows
- Game version 2.2.0

**Additional context**
When I chop a tree or a wooden door, sometimes I can't pick wood or the door up.
When I close X or Z menu, sometimes I randomly drop what I'm holding. I think something is wrong with dropping items.

BenCheung0422:
Currently, there is inventory size limit implemented. Please confirm whether your inventory is full. You cannot pickup more items when your inventory is full.

BenCheung0422:
Inventory limit is implemented since 2.2.0-dev1. This is likely the circumference when having full inventory. Maximum number of item stacks player can hold is 27.

### Analysis

**Crafting new item**
When the newly crafted item cannot be put into the player's inventory, the item is dropped from the player immediately. The situation is similar to Minecraft. When you crafted an item that cannot stack with other stacks in the inventory, the player would close the inventory to drop it, though the player can swap the holding item on the cursor and a stack in inventory, which is not possible in Minicraft+ at the moment. So the player can see an item dropped on the floor when the menu is closed.

**Opening menu**
When opening a menu, for example crafting menu or inventory menu, the item stack holding is supposed to be returned to the inventory. However, the inventory count does not count in holding stack, so when the inventory is full when returning held item stacks, they can just be dropped. When the menu is opened, the time is frozen, so the stack dropped can be seen when closing menu.

**Picking items up**
When the inventory is full, no more items can be picked up when they cannot stack with any stack in the inventory. Then, if the inventory is still full before the item entity disappearing, it is lost.

Inventory fullness is one of the mechanism that has to be noticed anytime to prevent such situation happening, as a part of gameplay.

If you can confirm the situation, I will convert this issue as a question.

PR #686, opened 2024-06-18
Add inventory capacity counter

BenCheung0422:
Self-explained by the title, this adds an inventory capacity counter, as demonstrated below:

https://github.com/MinicraftPlus/minicraft-plus-revived/assets/74168521/d497ec12-8e51-4e95-a854-e91ad1e1a369

The position of the counter is shifted upward to prevent potential collision of the counter with the title of the menu. Also, it is placed at the corners for convenience to users, instead of other places below blocking more sights. And there is a background to prevent similar colors causing confusion from the background.

Though there might be potential issues when the menus are too long, caused by long lengths, but should be able to be fixed by #543.

This can let players know when they have the inventories full. Potentially related to #685.


Litorom:
Looks good but could be better aesthetically. Try making it use the menu background. Similar to how durability/arrows worked in Eye’s hotbar rewrite.

BenCheung0422:
I have changed the design of the counter, minorly using similar techniques as in #581, as demonstrated below:

https://github.com/MinicraftPlus/minicraft-plus-revived/assets/74168521/d4abeeb1-8b22-4f2b-9960-988a5da40e67

Upon the design, the color of the text changes according to the fullness of the inventory while the shape of the counter changes when the width of the text changes. The position is lock at the corners and a little overflow is preferred (when focused) as of the limited area of the inventory menu (when empty, or shortest width). The counter is expanded with the limit when focused, so they will not collide with another counter on the other menu at all.

Litorom (approved):
The new look is interesting. I approve.

PR #688, opened 2024-06-20
Update some code handling strings where performance issues occur

BenCheung0422:
This improves a little performance issue. Here, `Double#parseDouble` is used to *validate whether the input is double*. It can be problematic as it throws exceptions when the input is invalid while the input is invalid most of the time. It is known that creation of exceptions are *expensive* due to the creations of stack traces. This pull request changes it to regex matching, though I doubt there would be an entry of number with localization enabled.

1~5% performance improvement is already significant when CPU time can be reduced. This kind of use cases (using methods with exceptions) will be carefully assessed when it can affect performance, in the code optimization task in the near future.

- When mainly browsing world:
image
- When browsing around menus and displays full of text:
image
- When testing #686 (mainly with inventory menus):
image
- When browsing around menus and displays full of text with this pull request:
image


BenCheung0422:
Furthermore, I have found another method with some performance issue (tested by browsing text-based displays):
image
Fixed by:
image

PR #691, opened 2024-07-24
Correct light overlay overworld opacity level

BenCheung0422:
The recent change to the lighting system has changed the mechanism to lighting. (#579) Originally it was linear constant value subtraction with a maximum value of 128 (in RGB) and clamping (negative results become 0). However that is not quite right to be in a lighting system, it is not how dimming should be done, so another kind of alpha composition is done.
I was not realized that lighting was not completely black when doing composition, so this corrects the opacity level when the player is on level >= 3. (Maximum opacity of black becomes 80% instead of 100%)

In the future, this lighting system will be further changed to use gamma adjustment and detailed alpha compositions, as in Minecraft, to tackle the issue of true coloring (without too much lightening). Also, since the supports to deeper color depth (10-bit) and HDR by monitors have increased along the time, there might be plans to support deeper coloring in game (texture sources might still be 8-bit) in the future.


BenCheung0422:
I have attached the demos before and after this change.

https://github.com/user-attachments/assets/4a1fb81b-6fe2-4131-ae71-13eb6a7c4796


https://github.com/user-attachments/assets/07e241c0-d17d-48d3-9e71-3d8f0c520e68

PR #697, opened 2024-08-28
Sync 2.2.X branch with main

BenCheung0422:


Makkkkus:
These branches are supposed to be behind main. They are working branches for eventual patches for these versions.

PR #722, opened 2024-11-28
Fix incorrect Russian letter ordering in font

### LINKED ISSUE #721: Incorrect letter order in russian localization
the letters "ju" and "ja" are in the wrong order in the font, which leads to confusion.

To reproduce:

1. Enter the game
2. Change the localization to russian
3. match the texts in game, for example "achievement", with the translator's translation, to see this mistake

definitely there is in versions 2.2.1-dev2 and 2.2.0

the first screenshot shows the in-game version , the second one is fixed, and the third one is the cause of the problem

Screenshot_20241128_092523
Screenshot_20241128_092534
Screenshot_20241128_092627

BenCheung0422:
Fix #721
A note that if there are resource packs that made changes to font, please be aware of this issue. They can still be fixed in older versions by just the correct ordering of letter glyphs in the sprite sheet.


Sernimator (commented):
Now the order of the letters is correct and they should be displayed correctly

Makkkkus (approved):

PR #726, opened 2024-12-04
Application Configuration for Gradle

zandgall:
Added the "Application" plugin to the client and server projects, which already use it to start their respective programs. Also, makes the client application run by default when executing the 'run' task in the base project.

Required to actually run Minicraft from terminal/command line alone*, which affects maybe no one other than me. But, also potentially makes it easier for people with different IDEs to begin developing without any config issues.

*`./gradlew run` or `./gradlew client:run` on linux, or `.\gradlew.bat run`/`.\gradlew.bat client:run` on windows


zandgall:
Haha I should explain my changes better then. The server can still be run via "./gradlew server:run", or however you intend to run the server in an IDE.

Before this change, both [server](https://github.com/MinicraftPlus/minicraft-plus-revived/blob/478c266719bfe0dafc92815424ebce532e871384/build.gradle#L74) and [client](https://github.com/MinicraftPlus/minicraft-plus-revived/blob/478c266719bfe0dafc92815424ebce532e871384/build.gradle#L61) had "application {}" objects, telling gradle how these projects can be executed. However attempting to do "gradle run" on either of these projects would fail, because the application plugin is only enabled for the [base project](https://github.com/MinicraftPlus/minicraft-plus-revived/blob/478c266719bfe0dafc92815424ebce532e871384/build.gradle#L2), and not for the subprojects. All this change does is _enable_ the client and server with the application plugin as they seem intended to be. Making the client run by default is optional, but server is still runnable.

This is mostly a peeve of mine because I _have to make these changes locally_ in order to run the game during development.

BenCheung0422:
I see your point, but I suggest removing the part that by default, the client one would run.

Makkkkus (dismissed):
Pretty simple, and easy. Fixes a problem I didn't know how to fix.

zandgall:
If wanted - you _can_ remove the client-by-default run command, just by removing lines 85-88 (...and maybe line 2, I'm not sure, I'll have to check if having plugin `id  "application"` with no `application {}` object builds and runs

zandgall:
So I confirmed that if you don't want client to run with `gradlew run` then you can just remove lines 85-88.

_However,_ README.md right now says that running `gradlew run` should run the game, so it seems that running client by default is at least intended behavior.

Makkkkus (approved):

PR #727, opened 2024-12-05
Split tile arrays into Chunks

zandgall:
Removes tile and data arrays from Level, and disperses it among Chunks. These are managed in ChunkManager, used by Level and LevelGen in order to access Chunks and Tiles.


BenCheung0422:
This task in fact should be a part of the engine rewrite, so I put this on hold.

src/client/java/minicraft/level/Chunk.java
```diff
@@ -0,0 +1,39 @@
+package minicraft.level;
+
+import minicraft.level.tile.Tile;
+import minicraft.level.tile.Tiles;
+
+public class Chunk {
+	public static final int SIZE = 32;
```
Makkkkus, line 7: Chunks should probably be much larger. Perhaps 64?

src/client/java/minicraft/level/Chunk.java
```diff
@@ -0,0 +1,39 @@
+package minicraft.level;
+
+import minicraft.level.tile.Tile;
+import minicraft.level.tile.Tiles;
+
+public class Chunk {
```
Makkkkus, line 6: This should probably be made into a struct, and integrated into the ChunkManager class. I know this makes working with chunks slightly harder, but it causes data to be stored closer in memory, which speeds up any interaction.

See: https://stackoverflow.com/questions/19385853/understanding-cache-friendly-data-oriented-objects-and-handles

Makkkkus (commented):
Neat implementation of chunks. I'm very positive this is a good implementation but I do have some comments on the object-oriented approach.

src/client/java/minicraft/level/Chunk.java
zandgall: Agreed, I was hoping to get feedback on this. Even 128 would be fine, but 64 is probably a good size. A 64x64 chunk would probably cover the whole screen which I think is a good metric, may be useful for potential chunk tricks like caching things or something like that. Not sure what intended use for chunks are being considered.

src/client/java/minicraft/level/Chunk.java
zandgall: That's a good point!

Had I made it in C or C++, I probably would have made it a struct by default. Java OOP energy is corrupting my brain...

Do you have recommendations or thoughts on how a struct-like data structure is best implemented in Java?

If memory structure is also being discussed, I would just throw out there the potential of sticking the (2-byte) short tile id and (2-byte) short data together as one (4-byte) "int tile" in the format 0xTTTTDDDD (with T being the Tile id, and D being the tile data) 
I would make this choice if I were working in C, but I'm not sure if this would be effective or even beneficial _at all_ in Java. But just putting the idea out there in case it has any benefits.

Makkkkus: Uncertain, I am myself more familiar with Rust and .NET when it comes to this sort of stuff.

I think the easiest method would be to make two array of arrays. One for tiles, and another for data. The outer array would contain each chunk, and the inner array would be the contents of the chunk. This skips the need for a struct completely, and provides simple lookup and editing.

The other approach would be a simple array of chunk structs, each conaining two arrays, tiles and data. This approach might be more user-friendly, but I am unsure how Java treats structs internally, and if this would have any benifit at all.

Makkkkus: > If memory structure is also being discussed, I would just throw out there the potential of sticking the (2-byte) short tile id and (2-byte) short data together as one (4-byte) "int tile" in the format 0xTTTTDDDD (with T being the Tile id, and D being the tile data) I would make this choice if I were working in C, but I'm not sure if this would be effective or even beneficial _at all_ in Java. But just putting the idea out there in case it has any benefits.

I have seen this been done other places in Minicraft+, and it would certainly have a benefit if the first option is to be chosen. I am unsure about the performance cost/benifit, but it should theoretically reduce the memory footprint to about half, which is significant.

zandgall: That's fair, I would be fine with either implementation honestly. But I do agree that Chunk can (even _should_) be dissolved into ChunkManager.

If the struct version can be done simple enough and easy enough to read, it would probably be the better solution just in terms of the source code.
But I would expect the 2d arrays to be the more 'efficient' solution, just harder to understand what's going on by just looking at the code.

zandgall:
@Makkkkus I went ahead and tried having the tile data arrays directly in ChunkManager. It seems simple enough to understand, I think it works as is for now. Can be made back into its own Chunk class (or essentially a struct) depending on changes to level format or data structure.

Makkkkus (dismissed):
Very good

zandgall:
This should be is ready!

src/client/java/minicraft/level/ChunkManager.java
Makkkkus, line 1: I'm unsure if this file has enough utility to justify being its own class. Overhead is probably minimal, but wouldn't it be easier to merge this with level?

zandgall: ChunkManager is utilized by both Level and LevelGen at the moment, as well as level saving + loading as a general level tile data format. Where Level stores entities and tiles and acts as an object updating them, ChunkManager deals only with the tile data.

If you wanna merge it with level, that might be fine, idk how well that would work but that's how I saw it as its own separate thing. You could alternatively migrate more functionality/data from other level-related classes, and/or change its name to be less vague.

Makkkkus: I see, then I have no objections

Makkkkus (approved):

PR #729, opened 2024-12-15
Resourcepack Menu

zandgall:
Modified the resourcepack menu to have one list of toggleable options. Changes/adds a few language translations explaining how to use the new menu.


Makkkkus (approved):
Seems solid.

PR #732, opened 2025-01-07
Infdev

zandgall:
Extends terrain on all levels using the chunk system. Ask @TheBigEye if you want to know the practical limits, but creates 'infinite' levels.

- Levels are generated on a chunk-by-chunk basis
- Level generation (mostly) utilizes Simplex noise rather than sequential random generation
- Level save format is modified, Level*.miniplussave files are replaced with Level*/ folders containing `c.x.y.miniplussave` and `d.x.y.miniplussave` file for every chunk saved in the level (containing tile ids and extra tile data respectively)
- Entities are spawned in a specific radius around the player no matter where the player is

Things to consider
- Game stutters when chunks are generated. There is a 'stage' property of chunks that can be used to generate the chunks over multiple ticks, as long as generation functions are split into several stage generators
- Usage of `random.next____()` functions may cause chunks in a level with the same seed to generate a little differently (especially with structures?) depending on which chunks are generated before others.
- Cloud and dungeon levels still have boss fights positionally oriented around the middle of the level as determined by level size setting


zandgall:
This relies on #727

PR #741, opened 2025-04-06
Fixed Hardcore Death Menu Exploit

TheBigEye:
This fix addresses an exploit in Hardcore mode where, upon death, a player could continue playing in the same world. If the player manually saved the world before dying, they could simply choose the "Quit" option from the death menu to return to the main menu without saving, and then reload the world


Makkkkus (approved):
Removes the possibility to quit without saving on hardcore mode. Logical

PR #743, opened 2025-04-10
Fix paths returned by ResourcePack::getFiles

KalmeMarq:
In java 8 the client/game jar resource pack returns the directory paths with '/'.

It should only append '/' at the end when it's a directory path and the path doesn't end with '/'.


Makkkkus (approved):
Smart

PR #754, opened 2025-05-26
Use ItemStack with components for the instance properties

KalmeMarq:
Use an ItemStack to represent an Item instance, and make Item simply be a "definition"/type (as much as possible righ now). It could be named ItemInstance but visually ItemStack looks nicer.

Even though the ItemStack has a count and maxCount was moved to the Item class, I didn't remove the StackableItem because I didn't know what to rename it to. 

Eventually I would prefer to get rid of the whole NAME + _ + AMOUNT string usage and just use the ItemStack. But I don't know if it should be done in this pull request.


KalmeMarq:
Separate the definition/type from the instance. When/if a proper (dynamic) Registry is added the Item would be the value of it. The ItemInstance/ItemStack simples stores what Item it contains, count, and data (if any) of that particular instance. 

In terms of memory it would be more noticeable if there were items with a bit more "static"/definition properties of their own. Copying all that around would be a waste. I see it as more organize, it's what I mean.

Issue #746, opened 2025-04-19
Watering can particles in wrong location


Bumber64:
**Describe the bug**
Sometimes watering can particles spawn far away from the player's location.

**To Reproduce**
1. Fullscreen (haven't tested without)
2. Use watering can in various positions and directions

**Expected behavior**
Particles should only spawn where the player is watering.

**Screenshots**
![Image](https://github.com/user-attachments/assets/c2bd42e2-12c7-4227-906f-07bb60916c89)
Notice water droplets near tree to the right.

**Desktop**

- OS: Windows 8.1
- Game version: 2.3.0-dev2

**Additional context**
Fullscreen with desktop resolution 1920x1080. Image is resized 50%.

PR #760, opened 2025-07-26
Fix issue #746

Chan-Chan-Irish-Man:
[Issue #746](https://github.com/MinicraftPlus/minicraft-plus-revived/issues/746) brought up a bug where the water particles would appear in a seemingly random location whenever the watering can is used.

I tested this further and discovered the water particles:
* would always appear on collision objects such as trees and walls
* would usually appear south-east of the player, although if a collision object was directly south or east of the player, the particles would appear there
* randomly would appear near the player

This bug would happen regardless if the game was in full-screen or windowed mode.

My fixes are as follows:
* in WateringCanItem.java interactOn
  * px and py are now calculated based on player coords plus 16 pixels times the attackDir outside the for loop
  * these values are then added by a random.nextInt(5) - 2 pixel offset to create x/y
  * and px/py are added by random.nextInt(3) - 2 pixel offset
* in WaterParticle.java
  * vastly simplified the algorithm which now uses Integer.compare 

Possible improvements:

The 16 number is what I put in as I thought it looked the nicest, but it can be changed to a larger or smaller number, preferably to a named const, but it is up to you of course.


src/client/java/minicraft/item/WateringCanItem.java
```diff
@@ -57,16 +56,18 @@ public boolean interactOn(Tile tile, Level level, int xt, int yt, Player player,
...
+				int px = player.x + 16 * attackDir.getX();
+				int py = player.y + 16 * attackDir.getY();
+
 				for (int i = 0; i < 4; i++) {
 					SpriteLinker.LinkedSprite splash = spriteSplash[random.nextInt(spriteSplash.length)];
-					// 2-pixel deviation for centering particle sprites.
-					int destX = player.x - 2 + 12 * attackDir.getX() + random.nextInt(9) - 4;
-					int destY = player.y - 2 + 12 * attackDir.getY() + random.nextInt(9) - 4;
-					int x = player.x - 2 + 4 * attackDir.getX() + random.nextInt(5) - 2;
-					int y = player.y - 2 + 4 * attackDir.getY() + random.nextInt(5) - 2;
```
BenCheung0422, line 66: `-2` for all those lines is to *center* the sprite, which contains a 8x8 tile, while the particle has a smaller radius. On the other hand, this simulates the starting point at the place the item likely renders near the player character, with the ending point of a distance (8 entity coordinate units on average) further,

BenCheung0422 (commented):
Since I am the only author of those two files, I would like to clarify the initiatives and the purposes of those algorithms and code, for future references, reviews, or whatever.

src/client/java/minicraft/item/WateringCanItem.java
Chan-Chan-Irish-Man: I also reverted this change as the original splash radius does look better than the more concentrated stream my solution created.

